Issue #2417, opened 2015-01-19
WARN Task #8 for mcMMO v1.5.02-SNAPSHOT-b23 generated an exception


Sckaletz:
After updating to the latest Spigot build I'm receiving this error on all my servers with MCMMO installed.

http://pastebin.com/EbW91i9C

WhiteNightFury:
[12:14:12 WARN]: [mcMMO] Task #6 for mcMMO v1.5.02-SNAPSHOT-b23 generated an exception
	java.lang.NullPointerException
			at org.bukkit.craftbukkit.v1_8_R1.scoreboard.CraftCriteria.getFromNMS(CraftCriteria.java:42) ~[craftbukkit-1.8-R0.1-SNAPSHOT.jar:git-Bukkit-00eba53]
			at org.bukkit.craftbukkit.v1_8_R1.scoreboard.CraftObjective.<init>(CraftObjective.java:19) ~[craftbukkit-1.8-R0.1-SNAPSHOT.jar:git-Bukkit-00eba53]
			at org.bukkit.craftbukkit.v1_8_R1.scoreboard.CraftScoreboard.getObjective(CraftScoreboard.java:40) ~[craftbukkit-1.8-R0.1-SNAPSHOT.jar:git-Bukkit-00eba53]
			at com.gmail.nossr50.util.scoreboards.ScoreboardManager.getPowerLevelObjective(ScoreboardManager.java:373) ~[?:?]
			at com.gmail.nossr50.util.scoreboards.ScoreboardManager.powerLevelHeartbeat(ScoreboardManager.java:336) ~[?:?]
			at com.gmail.nossr50.runnables.player.PowerLevelUpdatingTask.run(PowerLevelUpdatingTask.java:10) ~[?:?]
			at org.bukkit.craftbukkit.v1_8_R1.scheduler.CraftTask.run(CraftTask.java:53) ~[craftbukkit-1.8-R0.1-SNAPSHOT.jar:git-Bukkit-00eba53]
			at org.bukkit.craftbukkit.v1_8_R1.scheduler.CraftScheduler.mainThreadHeartbeat(CraftScheduler.java:349) [craftbukkit-1.8-R0.1-SNAPSHOT.jar:git-Bukkit-00eba53]
			at net.minecraft.server.v1_8_R1.MinecraftServer.z(MinecraftServer.java:653) [craftbukkit-1.8-R0.1-SNAPSHOT.jar:git-Bukkit-00eba53]
[...]

PR #2423, opened 2015-01-20
Fix issue #2417

QuarkTheAwesome:
Fixed the NPE on Spigot builds pointed out in #2417.


t00thpick1:
This was a spigot issue, I don't like throwing hacky workarounds into the codebase sorry :(

PR #2427, opened 2015-01-23
repair-event

PowerOfGermany:
Hi Guys,

Actually I try to dopple or tripple the costs for your skill /repair.
For Example for each enchantment you have to pay one more diamond (or the other used materials).
But at the moment the repair event isn't included in the API or a possibility to change it via config.

So my request is:
Could you add the repair-event to the API or make it possibly to change it via config?

KR,
PoG


TfT-02:
This is a not a pull request, please reopen a new ticket with this request.

Issue #2419, opened 2015-01-19
Ore dupe bug with pistons and slime blocks with mining double drop.


mibby:
Moved to https://trello.com/c/QtelmXlx (dev only)

TfT-02:
Apologies for the late response, if this bug is back - we need to handle it asap.

mibby:
Great. I'll be here to test whenever an attempt to fix it is made.

ulumulu1510:
This bug is back, I can confirm it for the version that I compiled when pulling it from github yesterday.

Our server made a bug-report for our main Tech-administrator that can be viewed here:
https://docs.google.com/document/d/1r5NVBfvAo8dZFhYEWuFxjFdiqzqbXxeOUhoFIfr5LYY/edit#

I hope it is of help for you  and we can get the bug fixed as asoon as possible.

ulumulu1510:
I think I actually might have found the bug:

https://github.com/mcMMO-Dev/mcMMO/blob/master/src/main/java/com/gmail/nossr50/runnables/StickyPistonTrackerTask.java#L34

 This line should be:
mcMMO.getPlaceStore().setTrue(movedBlock.getRelative(direction));

Instead of:
mcMMO.getPlaceStore().setTrue(block.getRelative(direction));

MorroZ:
any changes yet?

ulumulu1510:
@MorroZ , the PR has been pulled. In a next version, it will be fixed I assume.

k3ldon:
still works...

TfT-02:
@k3ldon Version number?

MorroZ:
@TfT-02 w8 is there already a version where this bug is fixed? where can i find it?^^

k3ldon:
latest... I actually update for that reason
[mcMMO] Loading mcMMO v1.5.02-SNAPSHOT-b${BUILD_NUMBER}

Absentee23:
Can confirm, using build 33, the dupe still works.

TfT-02:
@Absentee23 With which Spigot build?

I've just tried to reproduce this bug and was not able to on git-Spigot-c3c767f-33d5de3 and mcMMO v1.5.02-SNAPSHOT-b34

MorroZ:
@TfT-02 

[...]

PR #2456, opened 2015-02-11
Resolve duplication Bug #2419

ulumulu1510:
With the current file, upon retracting the piston, the current block position gets unset and the block behind the piston is set to true. This causes a bug/ opportunity to duplicate.

With my PR, this issue get's resolved and the correct block (the block which is retracted) is being set to true.

https://github.com/mcMMO-Dev/mcMMO/issues/2419


magnusulf:
You are the best.

PR #2476, opened 2015-03-01
Fixed ArrayIndexOutOfBoundsException

ysl3000:
Fixed ArrayIndexOutOfBoundsException checking ListSize before getting first element.


t00thpick1:
This should never be <= 0  if hasMetadata() returns true, so this is unneccessary

PR #2489, opened 2015-03-11
Fixed duplication Bug #2419

isokissa3:
Fixed duplication bug, works on my test server


isokissa3:
Ok, I fixed my code  :+1:

PR #2576, opened 2015-05-28
Fix drops for stone types added in 1.8

### LINKED ISSUE #2466: New 1.8 Blocks are not supported fully
When mining, double drops do not give you the same item that you mined, but instead regular stone.

I am running the dev Build #31 (Feb 17, 2015 11:12:38 AM

astemus:
Fix: Makes new stone types double drop the correct stone when mined with silk touch.
Fixes #2466


astemus:
Fixes #2466

t00thpick1:
Fixed

PR #2694, opened 2015-09-26
Fixed NPE when skill is not in config

Maximvdw:
When a skill is not configured this would throw an NPE as the section is not found.

Found while trying to get information from an McMMOLevelUp event


Maximvdw:
Seeing:
https://github.com/mcMMO-Dev/mcMMO/blob/master/src/main/resources/config.yml
Default only a few skills are in the config. The tested skill was Excavation where it would throw this npe

Maximvdw:
bump?

t00thpick1:
Fix formatting to have explicit brackets.

PR #2709, opened 2015-11-12
Strip colours from party chat when logging to console.

JamieSinn:
Example of what it looks like with colours in it
image
This is so hard to read, it's just painful.


andrewkm:
Please add. Every time console logs show colors babies and kittens fall extremely ill.
If you must, make it a config value, but I don't see why anyone would ever want colors anyways.

PR #2726, opened 2015-12-12
If sender is console, strip colours for mctop

JamieSinn:
Ugly ugly colours.


andrewkm:
Do not pull. Still working on making it proper to remove end ESC uglyness.

andrewkm:
Ready to pull, tested and works. Nothing we can do about the ending "esc" but fixed all the other ones.

JamieSinn:
@t00thpick1

PR #2781, opened 2016-03-05
Fix infinite GreenThumb & ShroomThumb usage bug

Tslat:
Green Thumb and Shroom Thumb don't check for 0 or lower itemstacks, allowing players to continuously green or shroom thumb well into negative itemstack amounts provided they don't let go of the mouse button


t00thpick1:
Not sure if this is the correct way to fix this, but it is A way, and that will due in the meantime.

PR #2782, opened 2016-03-05
Require sneaking to activate abilities...

DorCoMaNdO:
Require sneaking to activate abilities if the player is holding an item with the offhand.

This also affects the Chimaera Wing and blast mining, but that seems like an ok effect since you wouldn't want to use the Chimaera Wing if you place a torch with the offhand while holding it or at some similar scenario.

Excuse me if I've done it wrong, it seems to work just fine, I haven't used Java in a really long while.


DorCoMaNdO:
Apparently right clicking air with a fist no longer raises the server side event if the player doesn't have anything in the offhand. The only way to fix it (that I could think of at least) is creating some sort of a "dummy" item, our own "air" if you will, for the offhand while the player has no item in the hand, that should restore the functionality.

PR #2918, opened 2016-04-25
Exploit fix, turning spawned pigs into zombie pigmans using lightning stike allows you gain XP

isokissa3:
Some servers have option to spawn lightning stike if conditions are meet and abusing this to turn spawned pigs to zombie pigmans allowing crazy xp gains. This could also be abused by an abusive staff member.

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/EntityListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/EntityListener.java
@@ -31,6 +31,7 @@ import org.bukkit.event.entity.EntityTameEvent;
 import org.bukkit.event.entity.EntityTargetEvent;
 import org.bukkit.event.entity.ExplosionPrimeEvent;
 import org.bukkit.event.entity.FoodLevelChangeEvent;
+import org.bukkit.event.entity.PigZapEvent;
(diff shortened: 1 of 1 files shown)


t00thpick1:
Can you fix the spacing to match our formatting?

isokissa3:
There you go! :)

t00thpick1:
I moreover meant tabs and spaces.  Each indentation block should be 4 spaces.   There should be no tab characters.

isokissa3:
Oh I see, I thought they ware using spaces already

t00thpick1:
Thanks

Tslat:
Tbh I feel like you'd be better off without this.

To get the conditions right to make this abuseable, you'd have to either:
1. Have a way to manually spawn lightning
2. Have abusive permissions/admins

If it's the former, then it should be up to the source of that functionality to support McMMO, not the other way around

If it's the latter, then it's already abuse that we shouldn't be catering to.
This would be akin to saying that all mobs spawned with commands or spawn eggs should be blocked from xp too, since they can be abused by admins

The way it is post-merge; natural lightning events no longer spawn pigmen that support xp. While a niche case, it's still a viable method.

PR #2964, opened 2016-06-20
Fix brewing stand fuel refill loop.

jjm223:
If a brewing stand had 1 as a fuel value, and a brew was started (setting the fuel level to 0), the brewing stand would refill on its own while mcMMO would keep setting the fuel value back to 0. This used up all of the available blaze powder.


jjm223:
Didn't see that a pull request for this issue already existed!

PR #3011, opened 2016-08-07
Correcting loop conditions related to "ghost block" mitigation.

SidShakal:
If radius is 1 (as it is when called elsewhere in the class), the loop
conditions fail to refresh some of the neighboring chunks, allowing for
ghost blocks to continue to occur on eastern and southern edges of the
chunk the player is in.

This commit fixes the loop conditions.

= Chunk diagram of old behavior =
```
-----
-RR--
-RP--
-----
-----
```

= Chunk diagram of new behavior =
```
-----
-RRR-
-RPR-
-RRR-
-----
```

= Legend =
P: the chunk the player is in
R and P: chunks that are refreshed
-: chunks that do not get refreshed


t00thpick1:
Good catch

PR #3077, opened 2016-12-07
Lagless Threaded Queue

gonzalociocca:
Hope to see this implemented, mcmmo has caused tons of lag on my factions having 150+ players connected at the same time, the time of each save from the users took 100ms each causing a huge lag spike of 5-10seconds on my server every 10 minutes.

The queue is tested, and thread safe, if the server is stopped the queue would be stopped also. (dunno if it can thrown an exception? when the server is closed of course.)


gonzalociocca:
bad of me, the runnable is sleeping with the thread.

PR #3095, opened 2016-12-29
fix #2979

### LINKED ISSUE #2979: [1.9.4] block placing issue
hello

I found some problems with place blocks.

first, place some blocks on center of the world.
http://i.imgur.com/6NKYLFZ.png
Diamond block X,Z (0,0)
Diamond ore X,Z (1,1) (1,-1) (-1,1) (-1,-1)

next, break a block around (0,0).
http://i.imgur.com/TVd77rX.png
For example, break (1,1) with pickaxe. Mining xp is NOT obtainable.

and, break other blocks around (0,0).
http://i.imgur.com/3e4vFoj.png
http://i.imgur.com/V9EkkiI.png
http://i.imgur.com/J1hqDtf.png
As of images, Mining xp is obtained. 

It can makes XP boosting,  duplicate ores with doubledrops.

server: spigot 1.9.4
mcmmo: 1.5.06-b145

kanesada2:
I fixed #2979.(the reporter of this is my friend.)
This is caused by that negative values of chunk coordinates start from -1, not -0.
For example, a block positioned at (-1, 0, 0) is in the chunk whose coordinates is **(-1, 0, 0)** actually, but mcMMO  regard this block as in the chunk **(0, 0, 0)**.
The reason of this behavior is that this plugin divide x or z or y value by 16 or 64 to get which chunk a block is positioned at.
So I simply change this to subtract 1 from that value if that is less than 0.

### DIFF
--- a/src/main/java/com/gmail/nossr50/util/blockmeta/HashChunkletManager.java
+++ b/src/main/java/com/gmail/nossr50/util/blockmeta/HashChunkletManager.java
@@ -187,6 +187,17 @@ public class HashChunkletManager implements ChunkletManager {
         int cx = x / 16;
         int cz = z / 16;
         int cy = y / 64;
+        
+        if(x < 0){
+            --cx;
+        }
+        if(z < 0){
+            --cz;
+        }
+        if(y < 0){
+            --cy;
+        }
+        
         String key = world.getName() + "," + cx + "," + cz + "," + cy;
 
         if (!store.containsKey(key)) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/gmail/nossr50/util/blockmeta/HashChunkletManager.java
```diff
@@ -187,6 +187,17 @@ public boolean isTrue(int x, int y, int z, World world) {
         int cx = x / 16;
         int cz = z / 16;
         int cy = y / 64;
+        
+        if(x < 0){
+            --cx;
```
Glitchfinder: Notably, this actually makes the problem worse, as it would result in a block with the coordinate -16 to be treated as -32. In other words, the lower-valued edges of any negative chunks are now being treated as the edges of the next lower chunk. That is much more easily exploited than the issue being patched was.

PR #3234, opened 2017-07-13
Re-added handlersList

FabioZumbi12:
Other events implementing really dont need handlerslist, but the class from where is implementing needs.


mcmonkey4eva:
^ Solid simple PR. Should be squash-pulled ASAP. 👍

PR #3235, opened 2017-07-15
Add new mobs to treasures.yml

Viltaria:


mibby:
```
[Server thread/WARN]: [mcMMO] Invalid material: RAW_RABBIT
```

As well as bat section is empty.

Issue #3249, opened 2017-08-02
Geen thumb beetroot permission


ezeiger92:
Plugin.yml gives permission for the beetroot item ("mcmmo.ability.herbalism.greenthumb.plants.beetroot"), but green thumb checks against the beetroot block.

The permission should be "mcmmo.ability.herbalism.greenthumb.plants.beetrootblock"

Issue #3250, opened 2017-08-03
Could not pass player fish event to McMMo (1.12.1 update)


MatheweB:
https://hastebin.com/iqefunabek.vbs

These are the errors I get when someone tries to fish in the new 1.12.1 update. A bugfixes would be appreciated! Otherwise, no errors found.

SXRWahrheit:
Cosign

https://hastebin.com/tisirigigo.vbs

mibby:
Confirmed as well. Pasting on github ticket so there is a copy here.

McMMO dev 196
Paper dev 1170 (Spigot 1.12.1)
```
[20:38:05] [Server thread/ERROR]: Could not pass event PlayerFishEvent to mcMMO v1.5.08-SNAPSHOT
java.lang.NoSuchMethodError: org.bukkit.entity.Player.getTargetBlock(Ljava/util/HashSet;I)Lorg/bukkit/block/Block;
	at com.gmail.nossr50.listeners.PlayerListener.onPlayerFishHighest(PlayerListener.java:243) ~[?:?]
	at com.destroystokyo.paper.event.executor.asm.generated.GeneratedEventExecutor9.execute(Unknown Source) ~[?:?]
	at org.bukkit.plugin.EventExecutor$1.execute(EventExecutor.java:44) ~[patched_1.12.1.jar:git-Paper-1170]
	at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:74) ~[patched_1.12.1.jar:git-Paper-1170]
	at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:62) ~[patched_1.12.1.jar:git-Paper-1170]
	at org.bukkit.plugin.SimplePluginManager.fireEvent(SimplePluginManager.java:514) ~[patched_1.12.1.jar:git-Paper-1170]
	at org.bukkit.plugin.SimplePluginManager.callEvent(SimplePluginManager.java:499) ~[patched_1.12.1.jar:git-Paper-1170]
	at net.minecraft.server.v1_12_R1.EntityFishingHook.j(EntityFishingHook.java:477) ~[patched_1.12.1.jar:git-Paper-1170]
	at net.minecraft.server.v1_12_R1.ItemFishingRod.a(ItemFishingRod.java:19) ~[patched_1.12.1.jar:git-Paper-1170]
[...]

Issue #3251, opened 2017-08-03
Default vanilla xp multiplier for fishing is 0


ezeiger92:
In FishingManager, the method getVanillaXpModifier returns 0 if your skill is too low for all tiers. I believe this is supposed to be 1, which would leave xp unaffected for low skill rather than removing it.

AllPlayed:
Hello
This will be fixed with the new 1.12.1 version of MCMMO, just be patient 🙂

ezeiger92:
I've haven't seen your name around before and don't see any contributions to mcmmo from you... Who are you?

AllPlayed:
I just want to help people 🙂

ezeiger92:
But who are you to say that anything will be fixed? If you're not involved with this project, you can't know that info.

AllPlayed:
Why not?

ezeiger92:
"I **know** this thing will be fixed despite knowing nothing about the project and having no authority over what is released"
Stop that, it's not helpful.

AllPlayed:
Hey, the developer tell everyone that this will be fixed on the new 1.12.1 version

ezeiger92:
You aren't the developer and you haven't provided links so I don't believe you.

PR #3252, opened 2017-08-05
Fix 3250

ezeiger92:
Fix for issue #3250, changed HashSet<Byte> to HashSet<Material> for calls to getTargetBlock.

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
@@ -240,7 +240,7 @@ public class PlayerListener implements Listener {
                 return;
 
             case IN_GROUND:
-                Block block = player.getTargetBlock((HashSet<Byte>) null, 100);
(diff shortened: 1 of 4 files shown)


SXRWahrheit:
Could you put a compiled version of this on Drive or something? Would really like to stop the error spam on my server before @t00thpick1 or @TfT-02 pull this :)

ezeiger92:
I'm lazy so here's the build we're using: it also fixes #3249 and #3251
Had to zip because github doesn't like jars in here
[mcMMO.zip](https://github.com/mcMMO-Dev/mcMMO/files/1209745/mcMMO.zip)

SXRWahrheit:
Great, thanks!

ezeiger92:
Just updated the zip because really stupid stuff with ice fishing

SXRWahrheit:
Does your new zip fix the Kraken? I was using your zip and still not getting Kraken.

PR #3297, opened 2017-09-18
Added McMMOPlayerLoadedEvent

games647:
The mcMMO player profile is loaded async. This means plugins that depend on mcMMO and needs the data directly after player join have to implement some waiting task which asks every x seconds if the account is loaded yet and aborts after y tries. This is unreliable and delays the operations the plugin can do on that data. 

Furthermore plugins won't notice if the player profile was reloaded, because for example the admin removed the profile (/mcremove) or changed the experience formula (/mcconvert) while the player is online.


games647:
Why was this closed? It's still relevant.

t00thpick1:
The area affected is being rewritten in 1.13 update

PR #3329, opened 2017-10-31
If the treasure is an enchanted book, enchant it.

hoorigan:
Added a bit of logic to check and see if the drop is an enchanted book, and if it is, force at least one enchantment to be applied.  As a warning, I'm not a programmer, so I could have incorrectly used the continue statement here.  This requires more experienced eyes and some testing, I'm sure.


hoorigan:
Also, if I mucked up how I submitted this, I apologize.  I've only really used git for small, personal projects.

hoorigan:
Redoing this because I'm a derp and fail at github.

PR #3330, opened 2017-10-31
Enchanted book fixes

hoorigan:
The aim of these changes is to make it possible to configure enchanted books as a type of fishing treasure, and guarantee that the enchanted books the player receives will have enchantments on them.


t00thpick1:
I'm not sure this is the right fix, but for now I guess it will do.

hoorigan:
I know, it's hacky.  This is what happens when sysadmins touch code.  :P

Thank you!

Issue #3359, opened 2017-12-20
herbalism don't level up


Ironic8b498:
I have trying to get this issue fixed since the end of October, when it was brought to my attention, when you farm carrots, potatoes, wheat, beets, netherwart, leaves and coco beans you don't get any XP, and the auto replant has about 5 second delay before it starts to work. I know herbalism works for sugarcane, grass, pumpkin, watermelon and cactus. Sorry if I missed testing some plants. My server is Spigot 1.12.2 tested this on latest DEV release #4

sigmaroot:
Duplicated by #3311

Glitchfinder:
Fixed by https://github.com/mcMMO-Dev/mcMMO/commit/6876f192de4b22f44d7080b65fbe71e68361824d.

Issue #3311, opened 2017-10-06
Herbalism not leveling UP


Ironic8b498:
if I don't have seeds, carrots, potatoes, beets, etc in my inventory it works sometimes once I pick up one of the items that is used for planting it stops leveling up, I'm running Spigot 1.12.2 and mcMMO-#208
https://youtu.be/snxFvGOEO6w

Glitchfinder:
This one was fun to track, and is a group effort. Fun fact: you can blame a combination of oversights by at least half the dev list for this particular bug, but it can actually be resolved by making use of the `org.bukkit.event.block.BlockGrowEvent` to unflag crops. Unfortunately, I don't have time to actually fix it right now, so if anyone, particularly @t00thpick1 gets some time, that would be nice.

PR #3384, opened 2018-01-24
Calculate experience gain before processing green thumb.

### LINKED ISSUE #3284: Herbalism - Greenthumb autoplant does not award XP
The Herbalism passive skill of greenthumb (auto planting a crop when harvesting) still fails to award you Herbalism XP.

Tested using the latest 201 dev build on Spigot 1.12.1.

ProgrammingAtDawn:
Should fix #3284 
Not sure whether it is related to #3359 or #3311

Basically before the change, XP was calculated looking at the newly planted block after green thumb instead of what was harvested.
Now first XP is calculated and then green thumb is taken into account.

### DIFF
--- a/src/main/java/com/gmail/nossr50/skills/herbalism/HerbalismManager.java
+++ b/src/main/java/com/gmail/nossr50/skills/herbalism/HerbalismManager.java
@@ -148,10 +148,6 @@ public class HerbalismManager extends SkillManager {
             }
         }
         else {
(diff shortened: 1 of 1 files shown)


Glitchfinder:
@t00thpick1 I can confirm that #3284 is an issue, but I don't currently have the ability to confirm whether this fixes it. On another herbalism related note, is it intentional that using bonemeal to finish crop growth prevents earning experience from that crop? If not, I should probably file a ticket for that.

Issue #3433, opened 2018-04-20
NullPointerExceptions on player matching commands


GiovanH:
Version: 
McMMO Jenkins Build #7 (Apr 15, 2018 2:18:08 PM), version 1.5.09-SNAPSHOT
Spigot 1.12-R0.1

How to produce: run any command that targets a specific player, like /inspect or /party invite

Stack Trace:
```[20:21:35] [Server thread/INFO]: GiovanH issued server command: /inspect GiovanH
[20:21:35] [Server thread/ERROR]: null
org.bukkit.command.CommandException: Unhandled exception executing command 'party' in plugin mcMMO v1.5.09-SNAPSHOT
	at org.bukkit.command.PluginCommand.execute(PluginCommand.java:46) ~[spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at org.bukkit.command.SimpleCommandMap.dispatch(SimpleCommandMap.java:141) ~[spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at org.bukkit.craftbukkit.v1_12_R1.CraftServer.dispatchCommand(CraftServer.java:641) ~[spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at net.minecraft.server.v1_12_R1.PlayerConnection.handleCommand(PlayerConnection.java:1385) [spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at net.minecraft.server.v1_12_R1.PlayerConnection.a(PlayerConnection.java:1220) [spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at net.minecraft.server.v1_12_R1.PacketPlayInChat.a(PacketPlayInChat.java:45) [spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at net.minecraft.server.v1_12_R1.PacketPlayInChat.a(PacketPlayInChat.java:1) [spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at net.minecraft.server.v1_12_R1.PlayerConnectionUtils$1.run(SourceFile:13) [spigotmc-1.12-R0.1.jar:git-Spigot-596221b-9a1fc1e]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_141]
[...]

PR #3440, opened 2018-04-30
Handle playerdata without bukkit.lastKnownName info

GiovanH:
Fixed a null pointer exception that would occur if the offlinePlayer object did not have a defined string for .getName();


GiovanH:
See https://github.com/mcMMO-Dev/mcMMO/issues/3433

Issue #3432, opened 2018-04-18
Disable repair/salvage use with /mcability


GiraffeCubed:
With /mcability toggled off, players can still accidentally salvage, or repair items and lose the items or enchantments. I have had players repair their items whilst fighting and lose their enchantments, the ability to toggle this off could resolve this problem.

It should be expected that toggling /mcability also affects repair and salvage.

Hopefully this is a simple request to implement!

SXRWahrheit:
This would be beneficial to me, also. I tried disabling salvage via a permission for a specific world today, but it doesn't appear to be working. I was able to block repair anvils, though.

It would be nice to not have to resort to blocking the feature.

t00thpick1:
Closing due to 1.13 overhaul changes in progress touching the affected areas.

SXRWahrheit:
Can you clarify?

PR #3451, opened 2018-05-10
Anvil toggle

GiovanH:
As per https://github.com/mcMMO-Dev/mcMMO/issues/3432

This patch makes anvils usable only when /mcability is toggled on. This keeps them inline with other skills.

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
@@ -422,12 +422,17 @@ public class PlayerListener implements Listener {
     @EventHandler(priority = EventPriority.LOWEST, ignoreCancelled = true)
     public void onPlayerInteractLowest(PlayerInteractEvent event) {
         Player player = event.getPlayer();
+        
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
```diff
@@ -423,11 +423,12 @@ public void onPlayerRespawn(PlayerRespawnEvent event) {
     public void onPlayerInteractLowest(PlayerInteractEvent event) {
         Player player = event.getPlayer();
 
-        if (event.getHand() != EquipmentSlot.HAND || !UserManager.hasPlayerDataKey(player) || player.getGameMode() == GameMode.CREATIVE) {
+        McMMOPlayer mcMMOPlayer = UserManager.getPlayer(player);
```
Glitchfinder: I won't be able to accept the commit with this line here. If you see the if statement you moved to below it, one of the conditions is that mcMMO has data for the player. There have been issues in the past where the plugin failed to load an existing player and overwrote them with a new player with empty skills. (Usually caused by using an SQL server hosted elsewhere.)

It would be advisable to just leave the initial check as is and then add a new check for the mcability flag afterward.

PR #3481, opened 2018-07-02
Mctop Cooldown message correction. Closes #3470

### LINKED ISSUE #3470: /mctop message broken.
`    Database:
        # Database command cooldown in milliseconds, cannot be changed below the default value. (1750)
        # Prevents players from spamming /mctop or /mcrank
        Player_Cooldown: 60000`

Changing from the default 1750 milliseconds to 60,000 (1 minute) still shoots the 1 second message when players try to use /mctop. The cooldown itself **does** indeed work, however the message is incorrect and doesn't follow the configuration file setting.

https://i.imgur.com/vae3jT8.png

JamieSinn:


andrewkm:
@t00thpick1 this doesn't seem like it fixed the problem.
Message still does not follow delay, instead in now simply spams with:
https://i.imgur.com/J0tT8Vo.png

PR #3503, opened 2018-07-26
Updated all the wooden tools in multiple files

thunderlight411:
changed the wood tools


thunderlight411:
Updated all the wooden tools in multiple files

PR #3551, opened 2018-08-22
update excavation on experience.yml

nemba246:
In 1.13, the name of the grass block changed from **grass** to **grass_block**. This caused no experience to be given while using a shovel on grass. This change was part of something mojang called the flattening and could possibly be the cause of some other experience related issues. 
https://minecraft.gamepedia.com/1.13/Flattening

### DIFF
--- a/src/main/resources/experience.yml
+++ b/src/main/resources/experience.yml
@@ -131,7 +131,7 @@ Experience:
         Dirt: 40
         Coarse_Dirt: 40
         Podzol: 40
-        Grass: 40
+        Grass_Block: 40
         Gravel: 40
         Mycel: 40
         Sand: 40


L3Technosmith:
I can confirm this fixes the issue for me.  Also in treasures.yml, "Grass" should be "Grass_Block" in the Excavation area.  I tested out those changes by setting the CAKE to 10% starting at level 0 and verified I got no cakes (nor xp bonus) in 100 grass, but got them at roughly 10% on dirt...then applied the fix, and got the cake with the grass block.

L3Technosmith:
I am sorry for not making my previous message more clear.  This wasnt fixed in the treasures.yml...in the Excavation section.

Momshroom:
Just noting that I was able to fix quartz ore not giving mining xp by
changing the name in the experience.yml to the new name nether_quartz_ore.

PR #3783, opened 2019-02-10
Changed witch xp multiplier to 1.0

Zed-I:
Changed the xp multiplier for witches from 0.1 to a more reasonable 1.0

### DIFF
--- a/src/main/resources/experience.yml
+++ b/src/main/resources/experience.yml
@@ -402,7 +402,7 @@ Experience:
             Magma_Cube: 2.0
             Ender_Dragon: 1.0
             Wither: 1.0
-            Witch: 0.1
+            Witch: 1.0
             Iron_Golem: 2.0
             Wither_Skeleton: 4.0
             Endermite: 2.0


nossr50:
they are 0.1 for a reason

because you can easily farm them without the use of mob spawners

Zed-I:
Every single mob is farmable. Search *mob name* farm and you can find a farm for whatever you are looking for. Witch farms are one of these farms that require a bit more effort than other efficient farms, you need to build them at a specific spot and clear out the surrounding area to get good rates.

Endermen still have a 1.0 multiplier and there are designs that doesn't involve endermite but still gives more spawns then a witch farm [https://www.youtube.com/watch?v=BwHS25jApB4](example)

Massively nefing monsters that can be farmed is not the solution. It will just give an advantage to those whe can figure out a clever way to farm something that hasnt gotten nerfed.

nossr50:
witch XP used to be 0, it had been 0 for like 5~ years
I made it 0.1 because I thought 0 was pretty lame

nossr50:
anyways, mcMMO is setup so most server admins don't have to worry about common exploits, you can change them in the config file yourself, that is why the config files exist, to give you the option

PR #3788, opened 2019-02-12
Herbalism compatibility with PlayerAnimationEvent

LogGits:
This change in theory should make McMMO herbalism compatible with plugins that utilise PlayerAnimationEvent. The main reason that I want to add this is to fix `no swing` false positives in `AAC` (though didn't get to test this). 

I would appreciate if someone could test this PR and verify that its working without any issues. (I am busy for the next few days so I haven't been able to test it. In theory it looks like it should work. I could have checked if the event `isCancelled()` however, this doesn't seem necessary quite yet. Though, if it is needed I can add it 👍 

Please don't merge this unless it's confirmed completely working. I will find some time to test it out within the next few days.

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
@@ -29,6 +29,8 @@ import com.gmail.nossr50.util.sounds.SoundManager;
 import com.gmail.nossr50.util.sounds.SoundType;
 import com.gmail.nossr50.worldguard.WorldGuardManager;
 import com.gmail.nossr50.worldguard.WorldGuardUtils;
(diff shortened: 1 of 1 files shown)


nossr50:
It's possible this would be a proper solution to the false positive by AAC or other similar anti-cheat plugins. I would like this to be tested as a fix before accepting the PR.

LogGits:
Ok so i tested this and it seems to be working as intended. No visible bugs caused by this, no swing false positive in AAC is resolved with this.

nossr50:
@LogGits Preferably I'd like to wait until 2.2 was playable and this was tested again. I understand you tested it in the master branch, I hope this is not too inconvenient for you.

PR #3793, opened 2019-02-15
Combine interact events and use respectful priority

FlyingPikachu:
Per the [Discord discussion](https://discordapp.com/channels/526933440214597677/527704381408411648/544711598028095490), this commit will join the two PlayerInteractEvents which were split in https://github.com/mcMMO-Dev/mcMMO-Classic/commit/76a987e1e0870c7a5a58e00df4b78b2b50cb082a#diff-c39b42b55a35a823cd763d795fe0b0e4

In a nutshell, the two methods are running different but related code at different event priorities. While this looks nice from a readability standpoint, it is not standard practice as it causes the following problems:

1. Other plugins cannot cancel non-critical mcMMO functionality such as skill activation because, contrary to popular belief, EventPriority.LOWEST is called first. Thus, it is impossible for, say, region protection plugins (which are actually _supposed_ to be LOWEST), to cancel Unarmed activation while selecting a region.
2.  Based on each event method's header comment, the original goal for splitting may have been to use one listener to "modify" the event and the other to "monitor" the event. However, this hasn't held up over time as nowadays Shroom Thumb can potentially cancel the event in the "monitor" listener.
3.  While some other events in this class are using EventPriority.MONITOR (called last) correctly for logging purposes, this is not the case for the interact event.

To remedy this, the two methods were joined into a single EventPriority.HIGH event. I selected this priority as it allows typical plugins at the default EventPriority.NORMAL to cancel skill activation, but also to avoid having to compete with plugins which (often mistakenly) use EventPriority.HIGHEST to cancel events. I've tested this commit to be effective when applied to the master branch and when used in conjunction with my plugin, Quests (listener class [here](https://github.com/FlyingPikachu/Quests/blob/master/src/main/java/me/blackvein/quests/listeners/PlayerListener.java)).

I've set `ignoreCancelled=true` so that cancelled events will be respected (even though the tag reads like "ignore cancelled events" it actually means "ignore events that have been cancelled"). Although I'm confident that there is no critical code being skipped here, if that does not hold true then `event.isCancelled()` checks [can be applied instead](https://bukkit.org/threads/ignore-cancelled-events.61155/).

Please see [the god post](https://bukkit.org/threads/getting-your-priorities-straight-the-plugin-version.788/) in reference to my priority claims. Assuming this commit is merged, I will submit an identical commit to mcMMO Classic as I feel developers will expect Classic to behave in a similar manner to the remake when building for both (as my plugin Quests intends to do).

### DIFF
--- a/bukkit/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
+++ b/bukkit/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
@@ -478,24 +478,27 @@ public class PlayerListener implements Listener {
     }
 
     /**
-     * Handle PlayerInteractEvents at the lowest priority.
+     * Handle PlayerInteractEvents at high priority.
      *
-     * @param event The event to modify
+     * @param event The event to watch
      */
-    @EventHandler(priority = EventPriority.LOWEST, ignoreCancelled = true)
-    public void onPlayerInteractLowest(PlayerInteractEvent event) {
+    @EventHandler(priority = EventPriority.HIGH, ignoreCancelled = true)
+    public void onPlayerInteract(PlayerInteractEvent event) {
         /* WORLD BLACKLIST CHECK */
-        if (WorldBlacklist.isWorldBlacklisted(event.getPlayer().getWorld()))
-            return;
+        if (WorldBlacklist.isWorldBlacklisted(event.getPlayer().getWorld())) {
+        	return;
+        }
 
         Player player = event.getPlayer();
 
         /* WORLD GUARD MAIN FLAG CHECK */
         if (WorldGuardUtils.isWorldGuardLoaded()) {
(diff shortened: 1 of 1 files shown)


FlyingPikachu:
@nossr50 Bump

nossr50:
@FlyingPikachu as I explained to you on discord, you need to PR request to the 1.12 branch

FlyingPikachu:
But I did?

branch

nossr50:
> But I did?
> 
> branch

I see that now, I feel uncomfortable accepting this PR at this time because I cannot test 2.2
If you submit this PR to master, I would not accept it there because I'm going to force push 1.12 to master in the future.

I would suggest waiting until 2.2 is playable and then resubmitting the PR

FlyingPikachu:
Well that's a pain, but alright.

nossr50:
I'm sorry for the inconvenience

PR #3844, opened 2019-03-30
German Locale Overhaul

OverCrave:
I've been working on an overhaul of the german locale (when better to do that than during the overhaul era). The current state of the german locale after my changes is the following:

- Most stuff should at least have SOME translation now, especially the party stuff had no translation at all for example.
- All the new overhaul strings have been translated as well.
- I changed every color/chat code in there from the [[]] style to the & style

Next I'll be looking at the consistency of the translations, their style, word-usage, etc. and most important grammar.


nossr50:
Thanks I appreciate it

PR #3856, opened 2019-04-04
German Locale Update

OverCrave:
- fixed inconsistencies with skill and ability names (for example some command names not having the same translation as the skill name)
- some grammar, punctuation and color/chat code work
- added the last still missing translations, should now be up to date with the current en_US locale (by 2.1.35)


nossr50:
thanks for the updated locale

PR #3890, opened 2019-04-18
Fix adminchat returning too early if send by the console

### LINKED ISSUE #3883: [BUG] Adminchat on Console
Using the adminchat command in the console does nothing at all unless there is at least two arguments.

"ac hi" -> no output or error
"ac hi player" -> works fine

OverCrave:
Fixes #3883 

"hasPlayerDataKey" was returning false too early if the console send only one argument. Putting it after each of the on/off argument checks and after the consoleUsage checks fixes this and properly either tells the concosole that it can't use the on/off arguments or sends the first argument as a chat message like it's supposed to.

Tested on a 1.13.2 spigot server.


nossr50:
Thanks for the PR I'll review it soon ❤️

PR #3891, opened 2019-04-19
Fixed axe salvage bug

Zed-I:
Changes max quantity to 3 from 2 on all axes since they are crafted with 3 resources not 2


nossr50:
Thanks

PR #3892, opened 2019-04-19
Fixed bug where you couldn't shake zombie/creeper heads

Zed-I:
Fixed bug where skeleton heads would drop in instead of zombie/creeper heads on successful shake

Creeper/zombie heads are now their own items and data is no longer used


nossr50:
Thanks!

PR #3897, opened 2019-04-20
Added excavation treasure to coarse dirt

Zed-I:
Added coarse dirt to all drops that dirt can drop.
Also added grass block to the quartz drop since there is no reason not to.


nossr50:
Thanks

PR #3900, opened 2019-04-24
Update locale_it.properties

Leomixer17:
Use a more correct form of the translation "taming" in the italian locale, that even prevents to show the three dots in the scoreboard.


nossr50:
Thanks

PR #3903, opened 2019-04-26
Wrong overfishing count when fishing first time

Zed-I:
If you fish in the same spot when you first log into a server, you will only be able to fish 3 times before getting the overfishing message.

A very minor bug, but a very easy fix.


nossr50:
The count starts at 1 and resets to 1

PR #3906, opened 2019-04-26
Added bamboo to the non-2 high plant blocks

Zed-I:


nossr50:
Would've made more sense for this to be part of the other PR but eh can't complain.

Zed-I:
Yeah definitely. I messed that up somehow, my bad

PR #3910, opened 2019-04-28
Try to load locale from plugin data folder first

mikroskeem:
~~I haven't tested this yet.~~ Got sick of replacing messages in jar so here's my attempt to make this situation better.

Feel free to add suggestions/whine/w.e, good night 😄 

This PR is ready and tested, message loading works properly and falls back to defaults from JAR sanely if anything fails.

### DIFF
--- a/src/main/java/com/gmail/nossr50/locale/LocaleLoader.java
+++ b/src/main/java/com/gmail/nossr50/locale/LocaleLoader.java
@@ -19,6 +19,7 @@ import java.util.logging.Level;
 public final class LocaleLoader {
     private static final String BUNDLE_ROOT = "com.gmail.nossr50.locale.locale";
     private static ResourceBundle bundle = null;
+    private static ResourceBundle filesystemBundle = null;
     private static ResourceBundle enBundle = null;
 
     private LocaleLoader() {};
@@ -37,6 +38,13 @@ public final class LocaleLoader {
     public static String getString(String key, Object... messageArguments) {
         if (bundle == null) {
             initialize();
+        }
+
+        if (filesystemBundle != null) {
+            try {
+                return getString(key, filesystemBundle, messageArguments);
+            }
+            catch (MissingResourceException ignored) {}
         }
 
         try {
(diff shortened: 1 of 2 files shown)


src/main/java/com/gmail/nossr50/locale/LocaleLoader.java
```diff
@@ -85,7 +91,12 @@ else if (myLocale.length >= 2) {
                 locale = new Locale(myLocale[0], myLocale[1]);
             }
 
-            bundle = ResourceBundle.getBundle(BUNDLE_ROOT, locale);
+            Path localePath = Paths.get(mcMMO.getMainDirectory(), "locale_" + (locale != null ? locale.toString() : "") + ".properties");
+            try(Reader localeReader = Files.newBufferedReader(localePath)) {
+                bundle = new PropertyResourceBundle(localeReader);
+            } catch (IOException e) {
+                bundle = ResourceBundle.getBundle(BUNDLE_ROOT, locale);
```
mikroskeem: Definitely should be logged. As I am not completely familiar with mcMMO codebase yet, I left it undone for now.

src/main/java/com/gmail/nossr50/locale/LocaleLoader.java
```diff
@@ -85,7 +91,12 @@ else if (myLocale.length >= 2) {
                 locale = new Locale(myLocale[0], myLocale[1]);
             }
 
-            bundle = ResourceBundle.getBundle(BUNDLE_ROOT, locale);
+            Path localePath = Paths.get(mcMMO.getMainDirectory(), "locale_" + (locale != null ? locale.toString() : "") + ".properties");
```
mikroskeem: `mcMMO.getMainDirectory()` has slash in the end, `Paths.get(...)` does not like that really.

mikroskeem: Also not sure how to handle `locale == null` right now, I need to look how `ResourceBundle.getBundle(...)` works and then skip attempting to load the locale from filesystem at all

mikroskeem: Since `mcMMO.getMainDirectory()` returns a path with slash in the end, I simply proceeded with string concatenation.

As `ResourceBundle.getBundle(...)` throws `NullPointerException` when `locale` parameter is `null`, then I added a check to ensure that `locale` is not null.

mikroskeem:
Example of mcMMO trying to load `chmod 000`-ed locale file.

image

Things continue to load fine after that.

nossr50:
I want you to change a few things in this PR and I'd accept it

1) don't load locales from the main plugin directory, do it from a subdirectory ("locale" would be fine)
2) copy out the en_us locale if it doesn't exist in said subdirectory
3) any missing keys should be loaded from the internal en_us locale as usual
then I'd hit accept PR

electronicboy:
> copy out the en_us locale if it doesn't exist in said subdirectory
And when a locale is updated?

mikroskeem:
>And when a locale is updated?

When it's deleted from locales folder?

nossr50:
After talking to catboy, perhaps it would be smart to not auto-copy locales into this folder, and leave it to a config option? That way only people who know what they are doing will mess around with it, and if I update any strings only people using custom locales will be on outdated strings.

electronicboy:
I wouldn't even say a config option, just a "if the file exists and the locale entry exists, use it, or go use the default", it has many advantages:

1) It's simple
2) Server owners need to only override what they care about, be that the whole translation or that one line in there which they don't like the sound of
3) Updates are hell, there is no real way to handle automatically updating peoples locale files unless you started versioning them and keeping track of every single change made

The only downfall is that I really see is informing people of its existence, but that just sounds like a wiki page down the road or something

nossr50:
Yeah I like those suggestions catboy

PR #3914, opened 2019-05-01

snake0053:
Add the japanese locale based on the latest locale_en_US.properties.
It is not perfect, but it should be better than current ja_JP.


nossr50:
よくできました

PR #3921, opened 2019-05-06
Update locale_zh_CN.properties

89009332:


nossr50:
You need to convert your Chinese characters into UTF16

89009332:
ok

nossr50:
> ok

Could you please correct the directory this file is located in?

Currently you have it in 

src/ main/ resources/ locale/

it needs to be in

src/main/resources/locale/

Your directory path includes spaces currently

PR #3928, opened 2019-05-15

snake0053:
Fixed the wrong translation and added the missing translation.
Translation compliant with version 2.1.60.


nossr50:
Thanks for the update

PR #3952, opened 2019-06-12

snake0053:
Added the locale added to mcMMO 2.1.76.

And I changed Alchemy's translation from `醸造` to `アルケミー`.
I thought that `醸造` was not suitable as a skill name.

If Nossor does not like `アルケミー`, close this PR.
I will PR for the corrected translation to `醸造` again.


nossr50:
Wouldn't 錬金術 fit better?

snake0053:
I also thought about that. However, I did not know how Nossor thinks, so chose 'アルケミー', which can be understood as English meaning.
If you think 'Alchemy' is the best, I will correct the translation to it😃

nossr50:
錬金術 is an accurate translation of Alchemy

Issue #3983, opened 2019-07-07
IronArmStyle BonusMax and BonusMin being ignored


doesnotskate:
At first, BonusMax and BonusMin were missing from advanced.yml so I added them as per the commented instructions, but no matter how much i tweak these settings, unarmed still deals 13 damage at max rank.

nossr50:
You won't be able to customize the damage for Iron Arm and many damage skills until 2.2, the settings are missing because they are gone. I'm going to be doing a sweep across all combat skills and exposing their various parameters to the config in 2.2.

CollinHerber:
@nossr50 This has been a problem for 6 years now, with unarmed being extremely overpowered resulting in the complete removal of the Iron Arm skill. Is there any sort of ETA on this? To see this issue persist from when I've hosted servers back in 1.8 to now is .. well... frustrating to say the least.

If there is anything I can do to assist let me know.

CollinHerber:
@doesnotskate I was able to find the area and build a solution that allows a level-incremented bonus damage system instead of a rank bonus system. 

My current formula yields: Every 10 level, you get a 0.05 increase with a max bonus damage of 5.
This significantly nerfs unarmed damage from the stock values and has a more balanced leveling of unarmed (IronArm) skill.

If you'd like this .jar with this formula let me know and I can provide it. If you'd like the formula to change slightly and a jar to match let me know, as I should be able to provide. 

I'm going to work on making this configurable hopefully soon and then open a PR to this repo with the updates so everyone can enjoy them.

nossr50:
I recently tested and tweaked all PVP skills quite heavily, during testing Unarmed was on the weaker end. With that said, there will be another round of balance testing in the future after 2.2

CollinHerber:
@nossr50 I have made unarmed completely configurable: https://github.com/mcMMO-Dev/mcMMO/pull/3993

Let me know if there is anything further you'd like me to do with this.

@doesnotskate Hopefully this gets merged and a new version is released soon with these implementations. While yes in 2.2 a full combat damage configuration is nice, this is a great way to get the feet wet and satisfy some open issues.

PR #3993, opened 2019-07-17
enhancement: Make unarmed ironarm damage configurable in multiple ways

CollinHerber:


CollinHerber:
Solves https://github.com/mcMMO-Dev/mcMMO/issues/3983

nossr50:
I don't agree with these changes, more or less I'd like to keep it in a rank based system. That is the reason for closing this PR.

CollinHerber:
@nossr50 There is a configurable option for using the ranking system. You could make that true by default should you so choose. Don't think it warrants a full close though.

nossr50:
I would rather that not be an option. You can come talk to me about it in discord to understand my thought process, but I'm not going to accept a PR just because you have good intentions.

I plan to make everything configurable, but I don't plan to make the rank system optional at this moment in time.

PR #4003, opened 2019-07-26
Updated Russian translation up to the last version

myfbone:
I went through all entire Russian lang file, added missing values and translated. Somewhere I also checked previous translation and corrected it. Added new developers to /mcmmo command (e.g. electroniccat)


nossr50:
@myfbone Thank you

PR #4027, opened 2019-08-21
Add additional info about horse speed in Beast Lore

QuantumToasted:
An additional string containing info about the horse's movement speed (in blocks per second) is now concat'd to the end of the Beast Lore info. Also, added a Combat.BeastLoreHorseSpeed locale string to service this info.

image


QuantumToasted:
Anyone who wants to test it, please let me know if this works on other types of horses such as skele/zombie horses. The code would likely suggest that it would work but I'd rather it be confirmed so I don't make a buffoon of myself on my first PR.

nossr50:
Thanks I'll review your changes shortly and if they are acceptable they will be pulled :)

PR #4029, opened 2019-08-22
Fix spelling error in Guilds.Salvage.Section.0

QuantumToasted:
"...use an gold block" -> "...use a gold block"
Thanks to Stabbin#1131 from the mcMMO Discord server for pointing this out, just wanted to make the PR on his behalf.


QuantumToasted:
I have no idea why there was a second line changed in the file, if it messes something else up, woops.

Issue #3918, opened 2019-05-01
[Request] Reached max level.


Remski01:
So, 

Could you add something what change the required xp to just `-1` or something else then what the next level actually requires in `/mcstats`as example.

As example, we have a cap level of 550. However, when you skill is `550` it still shows the amount of xp you need for the next level while they have reached the max level...? Its quite confussing in my opinion

I hope you can add or change something what can remove this.

Shrek5InTheatres2019:
Is implemented, please close :)

Remski01:
Thanks <3

PR #4041, opened 2019-08-30
Implementing request from issue #3918

Shrek5InTheatres2019:
Small change to indicate in /mcstats when a player is at their max level


nossr50:
I like this PR, I may make some style adjustments.

PR #4065, opened 2019-09-26
locale_lt_LT.properties

Vyciokazz:
I translated in Lithuanian language manny required info for player and player skils, in time i update this translation.


adamxp12:
Can one of the admins verify this patch?

PR #4097, opened 2020-01-07
Create Chat Switch Events

drewhoener:
Add events to monitor permanent switching of chat to be monitored for compatibility with other chat managers


drewhoener:
Don't know if I'm allowed to, but figured I'd bump this. 
It'd be incredibly helpful for a few things I'm developing and I don't believe there's any downsides besides the API rewrite incoming
If it's helpful I'd be happy to write these events in the api branch or wherever is best suited as well

PR #4100, opened 2020-01-10
Add PartyCreated and PartyDisbanded events

GaetanRAYNAUD:


electronicboy:
events are in the wrong place to be able to handle cancellation, not that you even handle that?

GaetanRAYNAUD:
> events are in the wrong place to be able to handle cancellation, not that you even handle that?

My idea is get the events after the party is created/disbanded, so putting them before would not allow me to do that.
But for sure I need to undo creation/disbanding of the party if the event is canceled. Or do you want me to send the event before and prevent if that append ?

electronicboy:
prevention is better then having to track everything that you just did and undo it, especially for things like disbanding

Worth noting that all events are going to be replaced/deprecated in 2.2, so not entirely 100% sure on if we should be adding more events at this phase either, but 2.2 is still a WIP....

GaetanRAYNAUD:
Ok, this request can be closed.
I'll wait for 2.2 and see what's inside.

PR #4175, opened 2020-05-01
Fix NPE issue with logger

Draycia:
I tested it, starts up fine, mcMMO placeholders work fine, etc.
Resolves issue a couple people had with the latest dev build.

I also replaced the other sysout since the comment next to it is no longer relevant.


nossr50:
I told you to put it in constructor :P
fixed in d023b8918

PR #4195, opened 2020-05-28
Don't ready hand when player can't use skill

Draycia:


Draycia:
Note that this is one of the checks used before actually activating skills, so it should be consistent with that.

mibby:
@nossr50 Is there any problems with this being merged in? It fixes an issue I have with actionbar notifications appearing in worlds where users do not have permission to use mcmmo.

nossr50:
Thanks

PR #4198, opened 2020-06-01
Clean up LimitedSizeList::add a bit

serg06:


TheBusyBiscuit:
The `LimitedSizeList` was removed in a recent commit to mcMMO, therefore I reckon that this PR can be closed :?

PR #4219, opened 2020-07-09
Update translation to french

Elikill58:
Add missing messages and comments
I tried to make all message with UTF16, but i'm not sure that all messages have been made.


nossr50:
Thanks

PR #4255, opened 2020-08-09
Delete locale_ko.properties

wonyoungb:
100% not working on latest version and this is old file 17 month ago


nossr50:
If it doesn't work submit an issue, if you want to update it then submit a PR.

PR #4290, opened 2020-10-05
Correcting references to the Potion of Leaping being made with Red Mu…

mldriscoll:
Correcting a couple of references to the Potion of Leaping being made with Red Mushrooms
- In the skill guide
- and in the list of concoctions ingredients (this one was stopping the leaping potions getting the brewing speed bonus)


nossr50:
Can you explain this in more detail? Why is this change needed and what does it solve? Alchemy is likely to get nuked, but I may accept the pull if I understand the issue better.

mldriscoll:
As I recall, there are two issues being addressed here:

- The brewing time reduction doesn't work when brewing a rabbit's foot (since it isn't in the ingredients list that the time reductions are triggered from) - this is the one I had to edit the potions.yml file for
- The in-game description lists the wrong ingredient (Red Mushroom instead of Rabbit's Foot), which confuses players who try to brew them

nossr50 (approved):
Thanks

PR #4296, opened 2020-10-14
FakeEvent checks

TheBusyBiscuit:
This PR adds a small quality of life improvement for developers, the `FakeEvent` interface.
I admit that there is no clear benefit from this other than making code like this:

```java
@EventHandler
public void onDamage(EntityDamageByEntityEvent event) {
  // or just 'ismcMMOEnabled' if you use a variable
  if (plugin.getServer().getPluginManager().isPluginEnabled("mcMMO") && event instanceof FakeEntityDamageByEntityEvent) {
    return;
  }

  // Logic
}
```

a bit cleaner since this is no longer event-specific and can be translated and re-used as a standalone method:

```java
@EventHandler
public void onDamage(EntityDamageByEntityEvent event) {
  if (isFake(event)) {
    return;
  }

  // Logic
}

private boolean isFake(Event event) {
  // or just 'ismcMMOEnabled' if you use a variable
  if (plugin.getServer().getPluginManager().isPluginEnabled("mcMMO") && EventUtils.isFakeEvent(event)) {
    return;
  }
}
```

The benefit of this being a super interface is also that there is lower maintenance than having a bunch of instanceof-checks in the `isFakeEvent` method.
While an empty interface can be a nuisance... there might be a few methods down the line that could see a use in there.

Otherwise I also made the `EventUtils` class final and added a private constructor, since this is a utility class with static methods, we don't wanna instantiate that anyway, so it's best to prevent it this way.
I also saw that mcMMO uses Jetbrains Annotations, however I couldn't see them being used in the other methods of that class, let me know if I should make use of them though nonetheless for this method.

### DIFF
--- a/src/main/java/com/gmail/nossr50/events/fake/FakeBlockBreakEvent.java
+++ b/src/main/java/com/gmail/nossr50/events/fake/FakeBlockBreakEvent.java
@@ -7,7 +7,7 @@ import org.bukkit.event.block.BlockBreakEvent;
 /**
  * Called when mcMMO breaks a block due to a special ability.
  */
-public class FakeBlockBreakEvent extends BlockBreakEvent {
+public class FakeBlockBreakEvent extends BlockBreakEvent implements FakeEvent {
     public FakeBlockBreakEvent(Block theBlock, Player player) {
         super(theBlock, player);
     }
--- a/src/main/java/com/gmail/nossr50/events/fake/FakeBlockDamageEvent.java
+++ b/src/main/java/com/gmail/nossr50/events/fake/FakeBlockDamageEvent.java
@@ -8,7 +8,7 @@ import org.bukkit.inventory.ItemStack;
 /**
  * Called when mcMMO damages a block due to a special ability.
  */
-public class FakeBlockDamageEvent extends BlockDamageEvent {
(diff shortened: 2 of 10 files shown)


nossr50:
This is a clever idea and implemented well, consider it approved

> I also saw that mcMMO uses Jetbrains Annotations, however I couldn't see them being used in the other methods of that class, let me know if I should make use of them though nonetheless for this method.

I have started to add them, but to do full coverage of the code base would be time consuming so it hasn't reached that point yet.

>Otherwise I also made the EventUtils class final and added a private constructor, since this is a utility class with static methods, we don't wanna instantiate that anyway, so it's best to prevent it this way.

In the config branch many classes which feature static getters/ets are being murdered because it was poor design, some utility classes will remain the way they are now though

TheBusyBiscuit:
> I have started to add them, but to do full coverage of the code base would be time consuming so it hasn't reached that point yet.

Yeah I can relate XD
I'll keep it in mind for the next PR though and maybe include the annotations for any new methods or so then.

> In the config branch many classes which feature static getters/ets are being murdered because it was poor design, some utility classes will remain the way they are now though

Ah cool, yeah static getters aren't too great.
I reckon classes like `EventUtils` will probably persist - as they don't acces any fields or so - but should the class ever disappear, then we can at least still do an `instanceof` check in this case.

PR #4298, opened 2020-10-16
Feature: asynchronous chunkmeta handling

FrankHeijden:
This PR aims to solve the synchronous writes mcMMO does when chunk meta needs to be written to the disk. Consistency between loads/unloads are guarded by `LockManager`, which ensures only one worker thread can access a chunk/block at a time.


nossr50:
I believe @t00thpick1 will review this when he finds the time, he is in the middle of a rework of this part of the code base and I will leave it to him.

PR #4305, opened 2020-10-21
Some small optimizations

TheBusyBiscuit:
## 1. ItemMeta
The main goal of this pull request was to reduce the calls to `ItemStack#getItemMeta()` because `getItemMeta()` creates a copy of the original `ItemMeta` everytime it is invoked. Reducing such calls can be quite beneficial to the performance as those throwaway copies take up memory and resources to create.
These copies can sometimes take quite a significant time to create for very complex Item Metas, like written books for example, where the copy reaches very deep into the ItemMeta's data.

There have been three occasions on which this line:
```java
if (item.getItemMeta() != null && item.getItemMeta().isUnbreakable()) {
```

could be reduced to
```java
ItemMeta meta = item.getItemMeta();

if (meta != null && meta.isUnbreakable()) {
```
which resulted in half as many ItemMeta copies for that particular method.
It is a small improvement, but an improvement nonetheless.
One one occasion I also updated the durability check to use the newer `Damageable` system over the old `getDurability()`.

## 2. Private constructors
On a couple of occasions I have added private constructors to methods with nothing but static methods to avoid instantiation of these classes.
The classes have also been made `final`.

## 3. api/exceptions
I added a generated `serialVersionUID` for the `IncompleteNamespacedKeyRegister` Exception, as it was still missing one.
I have also added some annotations to the parameters of exceptions (with parameters) since I feel like that's a good place to start with. After all, a `NullPointerException` **within** a thrown Exception would be really annoying XD

## 4. Standards
~~I tried to make this not become a reformatting PR, but while I was working on the aforementioned points, I did notice a lot of inconsistencies, so I tried my best to align some if/else structures that were in my way with the formatting standards outline in your [standards.md](https://github.com/mcMMO-Dev/mcMMO/blob/master/standards.md), if that is still to be followed.~~

Let me know if these are still up to date or if I should not get involved in any of the formatting here.


src/main/java/com/gmail/nossr50/skills/salvage/Salvage.java
```diff
@@ -15,7 +22,7 @@
     public static boolean arcaneSalvageDowngrades  = AdvancedConfig.getInstance().getArcaneSalvageEnchantDowngradeEnabled();
     public static boolean arcaneSalvageEnchantLoss = AdvancedConfig.getInstance().getArcaneSalvageEnchantLossEnabled();
 
-    protected static int calculateSalvageableAmount(short currentDurability, short maxDurability, int baseAmount) {
+    static int calculateSalvageableAmount(int currentDurability, short maxDurability, int baseAmount) {
```
nossr50, line 25: Why was the type changed?

src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java
```diff
@@ -317,8 +328,9 @@ private void dropTreeFellerLootFromBlocks(Set<BlockState> treeFellerBlocks) {
     }
 
     private int updateProcessedLogCount(int xp, int processedLogCount, int beforeXP) {
-        if(beforeXP != xp)
+        if (beforeXP != xp) {
```
nossr50, line 331: Try to avoid changing the style
 (I know mcMMO code style isn't consistent, its made more consistent in another branch)

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -183,11 +191,11 @@ public static void handleAbilitySpeedIncrease(Player player) {
...
             int ticks;
 
-            if(abilityLengthCap > 0)
-            {
+            if (abilityLengthCap > 0) {
                 ticks = PerksUtils.handleActivationPerks(player,  Math.min(abilityLengthCap, 2 + (mcMMOPlayer.getSkillLevel(skill) / abilityLengthVar)),
                         skill.getAbility().getMaxLength()) * Misc.TICK_CONVERSION_FACTOR;
-            } else {
+            } 
+            else {
```
nossr50, line 198: Avoid making style changes in a PR that is not targeting doing so.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -204,29 +212,29 @@ public static void removeAbilityBoostsFromInventory(@NotNull Player player) {
     }
 
     public static void removeAbilityBuff(@Nullable ItemStack itemStack) {
-        if(itemStack == null)
+        if (itemStack == null) {
```
nossr50, line 215: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -204,29 +212,29 @@ public static void removeAbilityBoostsFromInventory(@NotNull Player player) {
...
             return;
+        }
 
-        if(!ItemUtils.canBeSuperAbilityDigBoosted(itemStack))
+        if (!ItemUtils.canBeSuperAbilityDigBoosted(itemStack)) {
```
nossr50, line 219: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -204,29 +212,29 @@ public static void removeAbilityBoostsFromInventory(@NotNull Player player) {
...
             return;
+        }
 
 
-        //1.13.2+ will have persistent metadata for this itemStack
+        // 1.13.2+ will have persistent metadata for this itemStack
```
nossr50, line 224: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -204,29 +212,29 @@ public static void removeAbilityBoostsFromInventory(@NotNull Player player) {
...
         AbstractPersistentDataLayer compatLayer = mcMMO.getCompatibilityManager().getPersistentDataLayer();
 
-        if(compatLayer.isLegacyAbilityTool(itemStack)) {
+        if (compatLayer.isLegacyAbilityTool(itemStack)) {
```
nossr50, line 227: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -204,29 +212,29 @@ public static void removeAbilityBoostsFromInventory(@NotNull Player player) {
...
-            }
+            // This is safe to call without prior checks.
+            itemMeta.removeEnchant(Enchantment.DIG_SPEED);
 
             itemStack.setItemMeta(itemMeta);
             ItemUtils.removeAbilityLore(itemStack);
         }
 
-        if(compatLayer.isSuperAbilityBoosted(itemStack)) {
+        if (compatLayer.isSuperAbilityBoosted(itemStack)) {
```
nossr50, line 237: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -264,7 +274,7 @@ private static boolean isLocalizedSkill(String skillName) {
         return false;
     }
 
-    protected static Material getRepairAndSalvageItem(ItemStack inHand) {
+    static Material getRepairAndSalvageItem(ItemStack inHand) {
```
nossr50, line 277: My preference is that you explicitly declare public, private or static

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -298,26 +308,28 @@ public static int getRepairAndSalvageQuantities(ItemStack item) {
     public static int getRepairAndSalvageQuantities(Material itemMaterial, Material recipeMaterial) {
         int quantity = 0;
 
-        if(mcMMO.getMaterialMapStore().isNetheriteTool(itemMaterial) || mcMMO.getMaterialMapStore().isNetheriteArmor(itemMaterial)) {
+        if (mcMMO.getMaterialMapStore().isNetheriteTool(itemMaterial) || mcMMO.getMaterialMapStore().isNetheriteArmor(itemMaterial)) {
```
nossr50, line 311: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -298,26 +308,28 @@ public static int getRepairAndSalvageQuantities(ItemStack item) {
...
             //One netherite bar requires 4 netherite scraps
             return 4;
         }
 
-        for(Iterator<? extends Recipe> recipeIterator = Bukkit.getServer().recipeIterator(); recipeIterator.hasNext();) {
+        for (Iterator<? extends Recipe> recipeIterator = Bukkit.getServer().recipeIterator(); recipeIterator.hasNext();) {
```
nossr50, line 316: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -298,26 +308,28 @@ public static int getRepairAndSalvageQuantities(ItemStack item) {
...
             Recipe bukkitRecipe = recipeIterator.next();
 
-            if(bukkitRecipe.getResult().getType() != itemMaterial)
+            if (bukkitRecipe.getResult().getType() != itemMaterial) {
```
nossr50, line 319: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -298,26 +308,28 @@ public static int getRepairAndSalvageQuantities(ItemStack item) {
...
                 continue;
+            }
 
-            if(bukkitRecipe instanceof ShapelessRecipe) {
+            if (bukkitRecipe instanceof ShapelessRecipe) {
```
nossr50, line 323: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -298,26 +308,28 @@ public static int getRepairAndSalvageQuantities(ItemStack item) {
...
                 for (ItemStack ingredient : ((ShapelessRecipe) bukkitRecipe).getIngredientList()) {
                     if (ingredient != null
                             && (recipeMaterial == null || ingredient.getType() == recipeMaterial)
                             && (ingredient.getType() == recipeMaterial)) {
                         quantity += ingredient.getAmount();
                     }
                 }
-            } else if(bukkitRecipe instanceof ShapedRecipe) {
+            } 
+            else if (bukkitRecipe instanceof ShapedRecipe) {
```
nossr50, line 332: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -77,43 +109,40 @@ public void onEntityTransform(EntityTransformEvent event) {
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onEntityTargetEntity(EntityTargetLivingEntityEvent event) {
-        if(!ExperienceConfig.getInstance().isEndermanEndermiteFarmingPrevented())
+        if (!ExperienceConfig.getInstance().isEndermanEndermiteFarmingPrevented()) {
             return;
+        }
```
nossr50, line 114: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -77,43 +109,40 @@ public void onEntityTransform(EntityTransformEvent event) {
...
         }
 
         //Prevent entities from giving XP if they target endermite
-        if(event.getTarget() instanceof Endermite)
-        {
-            if(event.getEntity() instanceof Enderman) {
-                Enderman enderman = (Enderman) event.getEntity();
+        if (event.getTarget() instanceof Endermite 
+                && event.getEntity() instanceof Enderman) {
+            Enderman enderman = (Enderman) event.getEntity();
```
nossr50, line 124: I had them nested for readability, I don't like having too many statements within one if statement

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -77,43 +109,40 @@ public void onEntityTransform(EntityTransformEvent event) {
...
+                persistentDataLayer.flagMetadata(MobMetaFlagType.EXPLOITED_ENDERMEN, enderman);
             }
         }
     }
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onEntityShootBow(EntityShootBowEvent event) {
         /* WORLD BLACKLIST CHECK */
-        if(WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld()))
+        if (WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld())) {
```
nossr50, line 135: The space after an if shouldn't be there for style reasons

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -77,43 +109,40 @@ public void onEntityTransform(EntityTransformEvent event) {
...
             Player player = (Player) event.getEntity();
 
             /* WORLD GUARD MAIN FLAG CHECK */
-            if(WorldGuardUtils.isWorldGuardLoaded())
-            {
-                if(!WorldGuardManager.getInstance().hasMainFlag(player))
-                    return;
+            if (WorldGuardUtils.isWorldGuardLoaded() 
+                    && !WorldGuardManager.getInstance().hasMainFlag(player)) {
+                return;
```
nossr50, line 145: Instead of unnesting them it would be better to have a method do both checks, the goal is readability.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -138,31 +167,32 @@ public void onEntityShootBow(EntityShootBowEvent event) {
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onProjectileLaunch(ProjectileLaunchEvent event) {
         /* WORLD BLACKLIST CHECK */
-        if(WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld()))
+        if (WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld())) {
```
nossr50, line 170: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -138,31 +167,32 @@ public void onEntityShootBow(EntityShootBowEvent event) {
...
-                    return;
+            if (WorldGuardUtils.isWorldGuardLoaded() 
+                    && !WorldGuardManager.getInstance().hasMainFlag(player)) {
+                return;
             }
 
             Projectile projectile = event.getEntity();
 
-            if(!(projectile instanceof Arrow))
+            if (!(projectile instanceof Arrow)) {
```
nossr50, line 185: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -138,31 +167,32 @@ public void onEntityShootBow(EntityShootBowEvent event) {
...
                 return;
+            }
 
             projectile.setMetadata(mcMMO.bowForceKey, new FixedMetadataValue(pluginRef, 1.0));
             projectile.setMetadata(mcMMO.arrowDistanceKey, new FixedMetadataValue(pluginRef, projectile.getLocation()));
 
-            for(Enchantment enchantment : player.getInventory().getItemInMainHand().getEnchantments().keySet()) {
-                if(enchantment.getName().equalsIgnoreCase("piercing"))
+            for (Enchantment enchantment : player.getInventory().getItemInMainHand().getEnchantments().keySet()) {
+                if (enchantment.getKey().equals(NamespacedKey.minecraft("piercing"))) {
```
nossr50, line 193: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -180,8 +210,9 @@ public void onProjectileLaunch(ProjectileLaunchEvent event) {
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onEntityChangeBlock(EntityChangeBlockEvent event) {
         /* WORLD BLACKLIST CHECK */
-        if(WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld()))
+        if (WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld())) {
```
nossr50, line 213: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -276,21 +309,23 @@ else if (isTracked) {
...
             Player defender = (Player) event.getEntity();
 
-            if(event.getCombuster() instanceof Projectile) {
+            if (event.getCombuster() instanceof Projectile) {
                 Projectile projectile = (Projectile) event.getCombuster();
-                if(projectile.getShooter() instanceof Player) {
+                if (projectile.getShooter() instanceof Player) {
                     Player attacker = (Player) projectile.getShooter();
-                    if(checkParties(event, defender, attacker)) {
+                    if (checkParties(event, defender, attacker)) {
```
nossr50, line 321: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -276,21 +309,23 @@ else if (isTracked) {
 
     @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
     public void onEntityCombustByEntityEvent(EntityCombustByEntityEvent event) {
-        //Prevent players from setting fire to each other if they are in the same party
-        if(event.getEntity() instanceof Player) {
+        // Prevent players from setting fire to each other 
+        // if they are in the same party
```
nossr50, line 313: Use a block comment if you are going to break them into multiple lines

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -276,21 +309,23 @@ else if (isTracked) {
...
+                    if (checkParties(event, defender, attacker)) {
                         event.setCancelled(true);
                     }
                 }
-            } else if(event.getCombuster() instanceof Player) {
+            } 
+            else if(event.getCombuster() instanceof Player) {
                 Player attacker = (Player) event.getCombuster();
-                if(checkParties(event, defender, attacker)) {
+                if (checkParties(event, defender, attacker)) {
```
nossr50, line 328: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -313,29 +348,28 @@ public void onEntityDamageByEntity(EntityDamageByEntityEvent event) {
...
         Entity attacker = event.getDamager();
 
-        if(WorldGuardUtils.isWorldGuardLoaded())
-        {
-            if(attacker instanceof Player) {
+        if (WorldGuardUtils.isWorldGuardLoaded()) {
+            if (attacker instanceof Player) {
 
-                if(!WorldGuardManager.getInstance().hasMainFlag((Player) attacker)) {
+                if (!WorldGuardManager.getInstance().hasMainFlag((Player) attacker)) {
```
nossr50, line 354: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -313,29 +348,28 @@ public void onEntityDamageByEntity(EntityDamageByEntityEvent event) {
...
                     return;
                 }
 
-            } else if(attacker instanceof Projectile) {
+            } 
+            else if (attacker instanceof Projectile) {
```
nossr50, line 359: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -313,29 +348,28 @@ public void onEntityDamageByEntity(EntityDamageByEntityEvent event) {
...
 
                 Projectile projectile = (Projectile) attacker;
 
-                if(projectile.getShooter() instanceof Player) {
-                    if(!WorldGuardManager.getInstance().hasMainFlag((Player) projectile.getShooter())) {
-                        return;
-                    }
+                if (projectile.getShooter() instanceof Player 
+                        && !WorldGuardManager.getInstance().hasMainFlag((Player) projectile.getShooter())) {
```
nossr50, line 364: Nested if conditions are easier to read, if you want to join multiple if statements put them in a method to increase readability

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -313,29 +348,28 @@ public void onEntityDamageByEntity(EntityDamageByEntityEvent event) {
...
+                    return;
                 }
 
             }
         }
 
         /* WORLD BLACKLIST CHECK */
-        if(WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld())) {
+        if (WorldBlacklist.isWorldBlacklisted(event.getEntity().getWorld())) {
```
nossr50, line 372: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -388,15 +422,14 @@ else if (attacker instanceof TNTPrimed && defender instanceof Player) {
             Player attackingPlayer;
 
             //If the attacker is a Player or a projectile belonging to a player
-            if(attacker instanceof Projectile || attacker instanceof Player) {
-                if(attacker instanceof Projectile) {
+            if (attacker instanceof Projectile || attacker instanceof Player) {
+                if (attacker instanceof Projectile) {
                     Projectile projectile = (Projectile) attacker;
-                    if(((Projectile) attacker).getShooter() instanceof Player) {
+                    if (((Projectile) attacker).getShooter() instanceof Player) {
```
nossr50, line 428: Unwanted style change, see other comments.

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -414,19 +447,21 @@ else if (attacker instanceof TNTPrimed && defender instanceof Player) {
...
                 }
             }
         }
 
         //Required setup for processCombatAttack
-        if(attacker instanceof Projectile) {
+        if (attacker instanceof Projectile) {
             ProjectileSource shooter = ((Projectile) attacker).getShooter();
-            if(shooter instanceof LivingEntity) {
+            if (shooter instanceof LivingEntity) {
```
nossr50, line 464: Unwanted style change, see other comments.

nossr50 (changes requested):
Remove the style changes, if I bring them in it will cause a merge headache for me when I merge this branch down to configurable which has had a consistent style applied accross the whole codebase. 

Please try to have a minimal diff when you submit a PR.

src/main/java/com/gmail/nossr50/skills/salvage/Salvage.java
TheBusyBiscuit: The reason behind it was that Bukkit changed the type of durability from `short` to `int` quite a while ago.
So having it as a short, when it's actually an int seems a bit nonsensical.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
TheBusyBiscuit: The keword `protected` is generally redundant for static methods if the class cannot be instantiated.
`package-private` and `package-protected` are equivalent in such a scenario, hence why you could argue that the protected modifier would be redundant and causes confusion, since the principle of inheritance isn't applied here.

I can add `protected` if you want, but there are valid reasons for omitting it in cases like this to avoid confusion with the actual accessibility of the method.

TheBusyBiscuit:
Thanks for your feedback.
I already suspected something like that with the formatting, hence why I had kept it to a minimum, only affecting the already changed files. 
That's also why I asked for your opinion on this in the PR body, I wasn't sure if there was a huge style change coming up, I'm gonna try to refrain from formatting changes  if not explicitly targeted then :)
I hope I reverted all the changes back XD

src/main/java/com/gmail/nossr50/skills/salvage/Salvage.java
nossr50: Do you know when this was changed? mcMMO is currently supporting versions between 1.13.2 -> Latest

nossr50:
I'll take a look at your changes in the near future, but yeah configurable branch had a consistent style applied to the whole codebase, which means any diffs done on master are just added work for me when I go to merge, so that is why I want the diffs to be minimal.

src/main/java/com/gmail/nossr50/skills/salvage/Salvage.java
TheBusyBiscuit: This happened in the 1.13 Update, since Mojang removed meta/sub ids and everything became its own Material.
Since durability was just a fancy word for meta ids, it just happened to be a side effect of it but yeah, it's been an `int` since 1.13 and part of the `ItemMeta` now.
So no worries, this won't break anything, in fact since the project is built against 1.13 the previous `getDurability()` method which used to return a `short` also shows up with a deprecation warning.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
nossr50: In my comment I wasn't recommending that it be made protected, I was saying that I prefer the scope be explicitly defined. I don't know why it was originally protected, most likely it was a mistake.

src/main/java/com/gmail/nossr50/api/exceptions/McMMOPlayerNotFoundException.java
```diff
@@ -1,11 +1,13 @@
...
 
+import javax.annotation.Nonnull;
+
 import org.bukkit.entity.Player;
 
 public class McMMOPlayerNotFoundException extends RuntimeException {
     private static final long serialVersionUID = 761917904993202836L;
 
-    public McMMOPlayerNotFoundException(Player player) {
+    public McMMOPlayerNotFoundException(@Nonnull Player player) {
```
nossr50, line 10: We are not using javax annotations, please keep the null/nullable annotations consistent with the rest of the code base.
We are using jetbrains library for annotations. It's also a good idea to avoid using javax because I believe that is oracle specific and doesn't ship out with adoptopenjdk, I could be wrong though.

src/main/java/com/gmail/nossr50/chat/ChatManagerFactory.java
```diff
@@ -1,15 +1,27 @@
...
 
-    public static ChatManager getChatManager(Plugin plugin, ChatMode mode) {
+    /**
+     * This is a static utility class, therefore we don't want any instances of
+     * this class. Making the constructor private prevents accidents like that.
+     */
+    private ChatManagerFactory() {}
+    
+    @Nullable
+    public static ChatManager getChatManager(@Nonnull Plugin plugin, @Nonnull ChatMode mode) {
```
nossr50, line 24: Use the Jetbrains annotation library, not javax

nossr50 (changes requested):
Some minor changes are required

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
TheBusyBiscuit: Ah, got it. I reckon it was for internal purposes only or something like that.
Although I don't really see a reason not to expose this method.
I think this might as well be `public` then, you agree?

src/main/java/com/gmail/nossr50/api/exceptions/McMMOPlayerNotFoundException.java
TheBusyBiscuit: My bad with the annotations, I'll change them.
And javax is not oracle-specific, javax annotations are part of JSR305 from the Google Findbugs project, it is included in the spigot executable jar, so nothing platform-specific at all.

I took a look at the `pom.xml` tree and found this dependency to be inherited from `worldedit-core`, I usually exclude unwanted or conflicting libraries like that in the pom. I think it would help avoid mistakes like this in the future since jsr305 annotations aren't gonna be available during development at all.
What's your opinion about this?

This would be the relevant pom entry I would add to the worldedit dependency:
```xml
<!-- ... -->
            <exclusions>
                <exclusion>
                    <!-- We use jetbrains instead. Excluding this -->
                    <!-- prevents us from using inconsistent annotations -->
                    <groupId>com.google.code.findbugs</groupId>
                    <artifactId>jsr305</artifactId>
                </exclusion>
            </exclusions>
```

nossr50: You can add the exclusion

nossr50: I may have been thinking of JavaFX and not javax

nossr50 (approved):
Looks acceptable now, thanks for the PR.

PR #4308, opened 2020-10-23
Adding crossbows, shields, elytra, and tridents to default repair.vanilla.yml and fixing minor typo.

Momshroom:
Changed comment over netherite items to say "Netherite repairables" instead of "Diamond repairables"

Added Crossbow to string repairables
Added Shield to wood repairables, with oak planks as material
Added Elytra and Trident to Other repairables, using phantom membrane and prismarine crystals respectively.


omer03:
Might as well add warped fungus on a stick as well.

nossr50:
When T&C comes out I was planning to update the repair tables to include Tridents/Crossbows, with that said I'm not against pulling this in once I evaluate it the materials involved a bit more closely.

nossr50:
Thanks, this will do until the config update rolls out.

PR #4321, opened 2020-11-02
Two small fixes

TheBusyBiscuit:
## 1. Fixed `armorCount`
`#getArmorContents()` will always return an array of length 4, individual items are nullable though, so the snippet below will always result in `armorCount` being equal to four. As it should only count the actually equipped armorpieces though, this behaviour was corrected to include a null-check.

```java
for(ItemStack armorPiece : ((Player) target).getInventory().getArmorContents())
{
    armorCount++;
}
```

## 2. Piercing Enchantment
Enchantment names are deprecated since a while, as they are not only inconsistently named but may also not be guaranteed to remain unique in future versions.
That's why comparisons should rather use a NamespacedKey instead.
I changed a small bit in `EntityListener` for that, to look for a `minecraft:piercing` enchantment instead.


nossr50 (approved):
Looks fine, in the future place separate fixes into different PRs when possible.

PR #4328, opened 2020-11-08
Update japanese translation

ViaSnake:
- Update untranslated strings.
- Replace the old color code.
- Change the translations of some ability names.
- Replaces the old string with the new one.
- ~~Restore the string changed in #4275.~~ Sorry I was wrong.
  - ~~If make this string into Japanese: （コールオブザワイルド） {0}が{1}個必要です。~~
  - ~~{0} will show Bone and {1} will show the amount. The unchanged string is more appropriate for Japanese than the changed string.~~

**Change**
`src/main/resources/locale/locale_ja_JP.properties`


nossr50:
can you update this for the conflicts?

ViaSnake:
I don't have time right now. Therefore it is closed.
I'll make a new pull request when I have time.

PR #4340, opened 2020-11-24
Feature: Show the name of the party who's chat you're toggling into

Teriuihi:
This pr adds the Use_Channel_Names option to the chat.yml config. It's false by default.

When false the behavior remains the same as before and it simply says if you're joining or leaving Party/Admin chat. When true it uses the party name of the user who's toggling their chat channel. If the channel is the Admin chat it defaults to false since Admin doesn't have the Use_Channel_Names option.


nossr50:
Thanks I'll review the design choices of this PR and its code sometime soon!

nossr50:
I'm not sure if you meant to but you bundled a bunch of unwanted things in this PR as of your latest commits, a lot of it overlaps with rewrites I'm doing in Tridents & Crossbows (all of the party code is being rewritten) as well.

Issue #4343, opened 2020-11-28
Let Axes replant Cocoa


alpoi:
Allow other tools, specifically an axe, to trigger herbalism green thumb. Cocoa Beans cannot instabreak with a hoe, so using a hoe is inefficient.

nossr50:
This is a good suggestion, I'll see what I can do.

PR #4373, opened 2021-01-06
Issue #4343 Axe can replant Cocoa

Lyther:
As mentioned in Issue #4343 , I have added axe support to cocoa.
* If the player holds hoe or axe, `processGreenThumbPlants` is processed.
* If the player uses axe to break other ageable other than cocoa, the method returns false.


nossr50 (approved):
Looks good, thank you.

PR #4375, opened 2021-01-08
Updated german translations

TheBusyBiscuit:
This pull requests updates the german locale file (`locale_de.properties`) to include translations for all the newly added strings that were added since the last time this file was updated.

I made sure to convert any foreign characters into the respective UTF-16 correspondence.


nossr50:
Thank you

PR #4400, opened 2021-01-22
only add/remove scoreboards when teleporting if scoreboards are enabled

steve4744:
When teleporting in a blacklisted world, e.g. playing parkour and getting teleported back to a checkpoint, the parkour scoreboard is getting removed by mcmmo, even if scoreboards are disabled in the mcmmo config.

The PR adds a check that mcmmo scoreboards are enabled before attempting to add or remove a scoreboard.


nossr50 (approved):
Thank you

PR #4418, opened 2021-02-02
Update french translation

Elikill58:
Translate missing messages into french.

I suggest you website like https://crowdin.com/ which help with translation. Because, for me, it's difficult to find all messages which have been edited.


nossr50:
Thanks, the main reason we do not use crowdin is we do not qualify for a free open source license as we are a commercial open source project.

PR #4432, opened 2021-02-18
Added Maven workflow to validate pull requests and commits

TheBusyBiscuit:
As discussed on discord - I am making a pull request to add a GH Actions workflow which will automatically compile and test commits and pull requests.

I made sure to thoroughly document everything and outline which step does what.
The action only runs where necessary, if you prefer to have it only run on pull requests, the push section can just be removed from the workflow.

This also utilises the `actions/cache` action which will use the local .m2 Maven cache to speed up building.
For mcMMO this resulted in 2m7s of compile time vs. 3m30s.

The action will flag any pull request in case the proposed changes won't compile successfully or fail any of the unit tests.
It may speed up pull requests review in the future as broken changes can be easily spotted.
This also does not upload/distribute the resulting jar.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,40 @@
+# This workflow automatically tests new commits and pull requests as they come in.
+# Note that this does not upload any artifacts, you will need to compile mcMMO manually
+# if you wish to create the actual jar.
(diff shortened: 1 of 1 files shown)


nossr50:
Thank you

PR #4460, opened 2021-03-20
Add '.all' to all existing perk xp multipliers

Pabsb:
Added '.all' usage for all existing perk xp mulipliers to enable use of mcmmo.perks.xp.<xpboost>.all

**Note** 
I believe I did this right but if not criticism/comments would be very much appreciated and accepted


nossr50:
Looks good, thanks.

PR #4481, opened 2021-04-07
Small optimizations

TheBusyBiscuit:
## Optimized Permission checks
So I wrote a quick method which logs all mcmmo permission checks.

These lookups should be quick and it is quite common for these to be fired very frequently.
However some small-scale optimizations can not hurt, as it may make things run smoother in the long run.

Particularly `McMMOPlayer#hasReachedPowerLevelCap()` is responsible for these many lookups since it loops through all primary skill types and fires a permissions lookup for each.
There isn't really a way around it, however `McMMOPlayer#hasReachedLevelCap()` (which checks for a specific Primary Skill Type) invokes  `McMMOPlayer#hasReachedPowerLevelCap()` too, resulting in the above seen 6 lookups per primary skill type.

This code portion for example may be optimized.
```java
//Check if they've reached the power level cap just now
if (hasReachedPowerLevelCap()) {
    NotificationManager...
} else if (hasReachedLevelCap(primarySkillType)) {
    NotificationManager...
}

```
Since we are inside of an `else if` block here, it means that `McMMOPlayer#hasReachedPowerLevelCap` must have returned false.
Therefore, we can save ourselves this further computation without breaking anything.
This PR proposes a small optimization to reduce permission lookup calls which can stack up quite a lot depending on what other plugins are at play.

| Before | After |
| ------- | ----- |
|Before|After|
| 98 lookups (max) | 85 lookups (max) |
| 72 lookups (min) | 59 lookups (min) |

These numbers are not 100% accurate or representative, as the amount of lookups depends a lot on internal logic.

## Enum Maps
I also switched out a few HashMaps for EnumMaps as their keys were enums.
EnumMaps are optimized for enums specifically and since the underlying data structure for them is an Array, it saves a bit on memory. They are also slightly quicker to compute than a regular `HashMap` in some instances, though the computation time will often be roughly the same.
Still, a decent way to optimize the code a bit further.

*I also added some missing annotations too :eyes:*

### DIFF
--- a/src/main/java/com/gmail/nossr50/datatypes/player/McMMOPlayer.java
+++ b/src/main/java/com/gmail/nossr50/datatypes/player/McMMOPlayer.java
@@ -67,6 +67,7 @@ import org.bukkit.metadata.FixedMetadataValue;
 import org.bukkit.plugin.Plugin;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -120,7 +121,7 @@ public class McMMOPlayer implements Identified {
 
     private PrimarySkillType lastSkillShownScoreboard = PrimarySkillType.values()[0];
 
-    public McMMOPlayer(Player player, PlayerProfile profile) {
+    public McMMOPlayer(@NotNull Player player, @NotNull PlayerProfile profile) {
         this.playerName = player.getName();
         UUID uuid = player.getUniqueId();
         identity = Identity.identity(uuid);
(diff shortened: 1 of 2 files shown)


DartPower:
Cool! 😃

DartPower:
I like your Slimefun4, best plugin with best performance :)

TheBusyBiscuit:
> I like your Slimefun4, best plugin with best performance :)

"best performance" :LULW:
Not even close but thanks I guess?

This is the mcMMO repository though, so please stick to the topic of this pull request or mcMMO in general.

DartPower:
Sorry for this off topic :(
But I like your work.

nossr50:
The changes are good, however this particular class has been rewritten in T&C so merging this would only cause some headaches, I will however port the changes included in this (or you can do it if you feel like it) down to the T&C branch.

PR #4482, opened 2021-04-08
Fix death messages losing formatting

### LINKED ISSUE #4463: mcMMO breaks death messages for kills using renamed items
Tested with just Paper & mcMMO

mcMMO version: 2.1.181
Paper Version: 565 MC 1.16.5

How it should look:
image

You should also be able to hover over the item for details, like so:
image

How it looks like with mcMMO:
image

lexikiq:
Fixes #4463

Uses a simple fix that temporarily clears the name of the attacking mob for a tick. Specifically, it listens to EntityDamageByEntityEvent at the highest priority, and if the damaged entity is a player who will be killed by the damage, it clears the name using MobHealthDisplayUpdaterTask, then sets it back a tick later using MobHealthbarUtils.handleMobHealthbars.

A more robust fix that avoids the small downside of the health bar being missing for a tick would involve adding Paper as a dependency, which I tried to avoid. Though, I think it could still be done without breaking Spigot support by checking [PaperLib.isPaper()](https://github.com/PaperMC/PaperLib/blob/7d357b35dd12835e12e40db85e4d39b195454cce/src/main/java/io/papermc/lib/PaperLib.java#L259-L265). If this is preferred, let me know and I can implement that instead.

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
@@ -12,6 +12,7 @@ import com.gmail.nossr50.events.fake.FakePlayerAnimationEvent;
 import com.gmail.nossr50.locale.LocaleLoader;
 import com.gmail.nossr50.mcMMO;
 import com.gmail.nossr50.party.ShareHandler;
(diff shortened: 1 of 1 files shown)


nossr50:
Thanks, the hacky fix will do for now.

PR #4483, opened 2021-04-08
Fix deaths from skeletons showing health bars

lexikiq:
When a player's last damage cause is a projectile, it finds the shooter and attempts to fix its name in death messages. Also fixes a long-standing bug where deaths to skeletons would post "Player123 was shot by Arrow"


lexikiq:
Totally proud of myself for noticing and fixing a 9-year-old bug 😅

nossr50:
Looks good, thanks

PR #4494, opened 2021-04-16
Some patches

### LINKED ISSUE #4189: Treefeller conflict in bordering protected areas.
Hi, 

I've had a user come to me with [an issue with my plugin Towny,](https://github.com/TownyAdvanced/Towny/issues/4001) where Treefeller cancels all breaks if one fakeblockbreakevent is cancelled. I traced it back to [here](https://github.com/mcMMO-Dev/mcMMO/blob/899a0152badf00869048a3ebd60053ec9f9c2aa6/src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java#L273) where you've actually got it marked with a TODO: saying it should likely be a continue.

If it were a `continue;` then players who are able to break some of the tree in the Towny plot they're working in, would at least be able to harvest the part of the tree they would otherwise be able to break manually.

As a `break;` it gets completely stopped from treefelling by Towny, with the standard "You don't have permission to break" message.

Warriorrrr:
This PR adresses the following bugs:
- Fixed blocks being dropped from blast mining even if yield was set to 0.
- Tree feller not working entirely if one fake block break event is cancelled. (Closes #4189)
- Fixes no woodcutting xp being rewarded if a tree is too big while using tree feller.

Added a removeAbilityBuff as a just-in-case when activating super breaker or giga drill breaker, in the unlikely event that the tool is still somehow buffed with the higher efficiency level.

### DIFF
--- a/src/main/java/com/gmail/nossr50/database/FlatFileDatabaseManager.java
+++ b/src/main/java/com/gmail/nossr50/database/FlatFileDatabaseManager.java
@@ -493,6 +493,8 @@ public final class FlatFileDatabaseManager implements DatabaseManager {
             skills.put(skill, getPlayerRank(playerName, playerStatHash.get(skill)));
         }
 
+        skills.put(null, getPlayerRank(playerName, powerLevels));
+
         return skills;
     }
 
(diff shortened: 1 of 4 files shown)


nossr50:
The change to FlatFileDB wasn't mentioned, however I already implemented that change so it is whatever.
Thanks

PR #4498, opened 2021-04-19
Add McMMOEntityDamageByRuptureEvent

lexikiq:
Fires a (Fake)EntityDamageByEntityEvent for Rupture/Bleeding and ensures the event hasn't been cancelled and the damage value is positive. Allows other plugins to block the damage, such as players toggling on/off PVP or entering safe zones.

### DIFF
--- a/src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
+++ b/src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
@@ -1,33 +1,23 @@
 package com.gmail.nossr50.events.skills.rupture;
 
 import com.gmail.nossr50.datatypes.player.McMMOPlayer;
-import com.gmail.nossr50.events.fake.FakeEntityDamageByEntityEvent;
+import com.google.common.collect.ImmutableMap;
 import org.bukkit.entity.Entity;
-import org.bukkit.entity.Player;
-import org.bukkit.event.entity.EntityDamageEvent;
+import org.bukkit.event.entity.EntityDamageByEntityEvent;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.EnumMap;
 
-public class McMMOEntityDamageByRuptureEvent extends FakeEntityDamageByEntityEvent {
-	public McMMOEntityDamageByRuptureEvent(@NotNull Player damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
-		super(damager, damagee, cause, getDamageMap(damage));
-	}
+public class McMMOEntityDamageByRuptureEvent extends EntityDamageByEntityEvent {
+	private final McMMOPlayer mcMMODamager;
 
 	public McMMOEntityDamageByRuptureEvent(@NotNull McMMOPlayer damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
-		this(damager.getPlayer(), damagee, cause, damage);
-	}
-
-	private static Map<EntityDamageEvent.DamageModifier, Double> getDamageMap(double damage) {
-		Map<EntityDamageEvent.DamageModifier, Double> damageMap = new HashMap<>();
-		damageMap.put(EntityDamageEvent.DamageModifier.BASE, damage);
-		return damageMap;
+		super(damager.getPlayer(), damagee, cause, new EnumMap<>(ImmutableMap.of(DamageModifier.BASE, damage)), new EnumMap<>(ImmutableMap.of(DamageModifier.BASE, (o -> -0.0))));
+		this.mcMMODamager = damager;
 	}
 
 	@NotNull
-	@Override
-	public Player getDamager() {
-		return (Player) super.getDamager();
+	public McMMOPlayer getMcMMODamager() {
+		return mcMMODamager;
 	}
 }
--- a/src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
+++ b/src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
@@ -10,12 +10,12 @@ import org.jetbrains.annotations.NotNull;
 import java.util.HashMap;
 import java.util.Map;
 
-public class McMMOEntityDamageByRupture extends FakeEntityDamageByEntityEvent {
-	public McMMOEntityDamageByRupture(@NotNull Player damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
+public class McMMOEntityDamageByRuptureEvent extends FakeEntityDamageByEntityEvent {
+	public McMMOEntityDamageByRuptureEvent(@NotNull Player damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
 		super(damager, damagee, cause, getDamageMap(damage));
 	}
 
-	public McMMOEntityDamageByRupture(@NotNull McMMOPlayer damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
+	public McMMOEntityDamageByRuptureEvent(@NotNull McMMOPlayer damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
 		this(damager.getPlayer(), damagee, cause, damage);
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/gmail/nossr50/runnables/skills/RuptureTask.java
```diff
@@ -44,20 +46,32 @@ public void run() {
...
-                        double damagedHealth = healthBeforeRuptureIsApplied - calculateAdjustedTickDamage();
-
-                        if(damagedHealth <= 0) {
-                            mcMMO.p.getLogger().severe("DEBUG: Miscalculating Rupture tick damage");
-                        } else {
-                            targetEntity.setHealth(damagedHealth); //Hurt entity without the unwanted side effects of damage()
-                        }
-                    }
+                if(damageTickTracker < DAMAGE_TICK_INTERVAL)
+                    return;
```
nossr50, line 50: What's the point of changing the logic to something that achieves the same thing?

src/main/java/com/gmail/nossr50/runnables/skills/RuptureTask.java
```diff
@@ -44,20 +46,32 @@ public void run() {
...
+                //Ensure the event wasn't cancelled and damage is still greater than 0
+                double damage = event.getFinalDamage();
+                if (event.isCancelled() || damage <= 0)
+                    return;
+
+                ParticleEffectUtils.playBleedEffect(targetEntity); //Animate
+                double damagedHealth = healthBeforeRuptureIsApplied - damage;
+
+                if(damagedHealth <= 0) {
+                    mcMMO.p.getLogger().severe("DEBUG: Miscalculating Rupture tick damage");
```
nossr50, line 72: If the event can modify this then this should probably get removed

src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
```diff
@@ -10,12 +10,12 @@
 import java.util.HashMap;
 import java.util.Map;
 
-public class McMMOEntityDamageByRupture extends FakeEntityDamageByEntityEvent {
-	public McMMOEntityDamageByRupture(@NotNull Player damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
+public class McMMOEntityDamageByRuptureEvent extends FakeEntityDamageByEntityEvent {
+	public McMMOEntityDamageByRuptureEvent(@NotNull Player damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
```
nossr50, line 14: Should probably never construct it with Player over McMMOPlayer since McMMOPlayer should **always** be available

nossr50 (changes requested):

src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
```diff
@@ -0,0 +1,33 @@
...
+		Map<EntityDamageEvent.DamageModifier, Double> damageMap = new HashMap<>();
+		damageMap.put(EntityDamageEvent.DamageModifier.BASE, damage);
+		return damageMap;
+	}
+
+	@NotNull
+	@Override
+	public Player getDamager() {
+		return (Player) super.getDamager();
+	}
```
nossr50, line 32: Should have a getter for McMMOPlayer instead which has its own ::getPlayer method

nossr50: EDIT: Or in addition to

src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
```diff
@@ -0,0 +1,33 @@
...
+import com.gmail.nossr50.events.fake.FakeEntityDamageByEntityEvent;
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.Player;
+import org.bukkit.event.entity.EntityDamageEvent;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class McMMOEntityDamageByRuptureEvent extends FakeEntityDamageByEntityEvent {
```
nossr50, line 13: This is a big no-no, fake events are for things that don't actually happen. Don't use them for damage that will be real.

src/main/java/com/gmail/nossr50/runnables/skills/RuptureTask.java
```diff
@@ -44,20 +46,32 @@ public void run() {
...
-                    }
+                if(damageTickTracker < DAMAGE_TICK_INTERVAL)
+                    return;
+
+                damageTickTracker = 0; //Reset timer
+                double healthBeforeRuptureIsApplied = targetEntity.getHealth();
+
+                //Ensure victim has health
+                if(healthBeforeRuptureIsApplied <= 0.01)
+                    return;
```
nossr50, line 57: I don't understand why you took time to change all the logic to be the same logic but different, its just changing the code style for no particular reason? Weird thing to do in a PR.

nossr50 (changes requested):
Would prefer you didn't change code styling without a good reason (I'm open to it, but no reason given just makes me not understand it). Anyways if you make the changes I'd be glad to accept the PR.

src/main/java/com/gmail/nossr50/runnables/skills/RuptureTask.java
lexikiq: Sorry, force of habit in reducing unnecessary tabs, I can undo the code styling changes.

src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
lexikiq: ah, misunderstood what this was for then

src/main/java/com/gmail/nossr50/runnables/skills/RuptureTask.java
nossr50: Thanks!

src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
nossr50: It's all good!

nossr50: You may want to have this extend SubSkillEvent (going off memory here) we should have some kind of "template" event for sub skills already in mcMMO

lexikiq: I don't think I can do that? That'd be a class extending two classes

nossr50: You're about to remove the class it is extending, I was offering a suggestion of what it should be extending instead

src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
```diff
@@ -1,33 +1,23 @@
...
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.EnumMap;
 
-public class McMMOEntityDamageByRuptureEvent extends FakeEntityDamageByEntityEvent {
-	public McMMOEntityDamageByRuptureEvent(@NotNull Player damager, @NotNull Entity damagee, @NotNull DamageCause cause, double damage) {
-		super(damager, damagee, cause, getDamageMap(damage));
-	}
+public class McMMOEntityDamageByRuptureEvent extends EntityDamageByEntityEvent {
```
nossr50, line 11: Hmm, I'm not sure whether or not having it extend EntityDamageByEntityEvent is a good idea here, might cause unwanted behavior with other plugins.

nossr50, line 11: Well, one way to find out if this is a problem is to simply put it in and see what kind of complaints we get

nossr50 (changes requested):
I'm sure if it's wise to extend EntityDamageByEntityEvent, that would be the main thing giving me pause, however maybe its worth trying out, I guess I'm primarily worried about potential conflicts with anti-cheat stuff, we could change the damage type to custom which may solve that.

src/main/java/com/gmail/nossr50/events/skills/rupture/McMMOEntityDamageByRuptureEvent.java
nossr50: Hmm on second or third thought, this is probably fine if we change our damage type to custom.

nossr50:
Alright lets merge this and see what happens
Thanks for the work!

lexikiq:
happy to help :)

nossr50:
@lexikiq if you're on our discord I can give you a role for contributors ( we have a dev channel )

lexikiq:
i'm qixils#0493 :>

PR #4508, opened 2021-05-01
Disable launching players from Fishing Anti-AFK

lexikiq:
The minecart flight issue is patched as of 20w14a

https://bugs.mojang.com/browse/MC-176387

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
@@ -794,7 +794,6 @@ public class PlayerListener implements Listener {
                 if(player.isInsideVehicle() && (player.getVehicle() instanceof Minecart || player.getVehicle() instanceof PoweredMinecart))
                 {
                     player.getVehicle().eject();
-                    player.setVelocity(player.getEyeLocation().getDirection().multiply(10));
                 }
 
                 //mcMMOPlayer.getFishingManager().setFishingRodCastTimestamp();


lexikiq:
hmm.. I guess I'll add a version check
image

lexikiq:
I have set the patch to only enable on 1.16+ (hope I used the CompatibilityManager right) and to not launch players for using the exploit. My rationale is explained in the commit's comments.

It's worth noting that this patch is ineffective at best as [vehicles other than minecarts](https://youtu.be/cln5mrh0zDc?t=590) can be used for this exploit as well, but that's outside the scope of this PR...

src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
```diff
@@ -789,12 +789,12 @@ public void onPlayerInteractMonitor(PlayerInteractEvent event) {
         if(event.getAction() == Action.RIGHT_CLICK_BLOCK || event.getAction() == Action.RIGHT_CLICK_AIR)
         {
             if(ExperienceConfig.getInstance().isFishingExploitingPrevented()
-                       && (heldItem.getType() == Material.FISHING_ROD || player.getInventory().getItemInOffHand().getType() == Material.FISHING_ROD))
+                       && (heldItem.getType() == Material.FISHING_ROD || player.getInventory().getItemInOffHand().getType() == Material.FISHING_ROD)
+                       && !mcMMO.getCompatibilityManager().hasPatchedMinecartFlight())
             {
                 if(player.isInsideVehicle() && (player.getVehicle() instanceof Minecart || player.getVehicle() instanceof PoweredMinecart))
```
TheBusyBiscuit, line 795: `PoweredMinecart` is already an extension of `Minecart`. So this check can be simplified.

Also, this may just be me but I feel like the Minecart Flight patch check should go into the if-statement with the Minecart check, not the one with the fishing rod.
It just seems like it makes more sense to me that way.

```suggestion
                       && (heldItem.getType() == Material.FISHING_ROD || player.getInventory().getItemInOffHand().getType() == Material.FISHING_ROD))
            {
                if (player.isInsideVehicle() && player.getVehicle() instanceof Minecart
                    && !mcMMO.getCompatibilityManager().hasPatchedMinecartFlight())
```

src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
```diff
@@ -789,12 +789,12 @@ public void onPlayerInteractMonitor(PlayerInteractEvent event) {
...
                 {
                     player.getVehicle().eject();
-                    player.setVelocity(player.getEyeLocation().getDirection().multiply(10));
```
TheBusyBiscuit, line 797: I didn't really understand why this is here or why it was removed until I saw your commit message.
Perhaps adding a comment about it here may be a good idea.

TheBusyBiscuit (changes requested):
Just some things that could be improved in my opinion.
But that may just be me

src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
lexikiq: Not sure why I should add a comment for removed code? Isn't that what version control/commit history is for?

TheBusyBiscuit: Completely fair point.
I was just pointing out my confusion and thought that commenting the line out and adding an explanation right next to it would make it easier to comprehend.

Though I don't have any say here anyway, it was just a suggestion from me which you can feel free to ignore if you want :)

nossr50:
The code here was never to patch a flight exploit, it was to prevent getting around anti-afk mechanics for fishing. Which is why I hurled players out of their minecart.

I'm willing to reopen this if there is some other reason to have this.

lexikiq:
Please reconsider. By hurling players out of their minecarts, without a configurable option at that, you allow players to rocket up into the air exactly like a Trident for completely free. This is a massive exploit that numerous players on our server had been abusing until I modded our server to find and fix the cause. Anti AFK mechanisms don't need to chuck players, you can just dismount them from the minecart and kick them if they remount too much.

nossr50:
@lexikiq Thanks for the follow up explanation

nossr50:
@lexikiq Does this only affect version 1.16? Why the version check?

lexikiq:
The version check was under the assumption that the code was trying to fix a vanilla bug which was patched in 1.16

nossr50:
@lexikiq Can you remove the version check and just change the PR to remove the ejecting players out of minecarts completely?

lexikiq:
@nossr50 Done

nossr50:
Thank you

PR #4515, opened 2021-05-08
Fix piston movement setting all moved blocks as non-new

levtey:
Mostly an issue for skyblock servers, the current protection stores all blocks moved by pistons as non-new. This, however, will only move the blocks that were already stored, allowing for pistons to move new ore blocks but stopping duplication.


Geolykt:
You might want to check your indentation settings, it isn't uniform with the previous code as I have spotted a few tabs in an otherwise all-spaces code

t00thpick1:
I'd still want to see extensive testing of your code to make sure it doesn't introduce things.

nossr50:
I'm in agreement with @t00thpick1 , unless this was rigorously tested on all MC versions it would have to be optional off by default type setting.

PR #4536, opened 2021-05-27
Fix subskill chat click bug

Code-Log:
When viewing skill info. The subskill click listener doesn't work. I fixed it by simply not allowing the plugin to reformat the subskill id. The function getNiceNameNoSpaces(...) doesn't produce a name that will be found by /mmoinfo

### DIFF
--- a/src/main/java/com/gmail/nossr50/util/text/TextComponentFactory.java
+++ b/src/main/java/com/gmail/nossr50/util/text/TextComponentFactory.java
@@ -286,7 +286,7 @@ public class TextComponentFactory {
             else
                 textComponent = Component.text().content(LocaleLoader.getString("JSON.Hover.SkillName", skillName));
 
-            textComponent.clickEvent(ClickEvent.runCommand("/mmoinfo " + subSkillType.getNiceNameNoSpaces(subSkillType)));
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit (commented):
Shouldn't `/mmoinfo` rather be fixed then instead of this workaround?

I gave a quick look at it and seems to perform the same operation to check for it:
https://github.com/mcMMO-Dev/mcMMO/blob/master/src/main/java/com/gmail/nossr50/commands/skills/MmoInfoCommand.java#L80

However, it seems to be gated behind a null check against this:
https://github.com/mcMMO-Dev/mcMMO/blob/0d2f370185d8f6a0ce86bb857fa0ed46742950e8/src/main/java/com/gmail/nossr50/listeners/InteractionManager.java#L64-L67

I feel like this should perhaps be tested/investigated first before the behaviour is changed here, since there is a chance that the intended behaviour could be restored :?

Code-Log:
Yeah, I guess mmoinfo should rather be fixed, but I don't know the code and this is just a quick fix that worked for my server.

On the off chance that someone wants this code, I'll close the PR, but leave my repo active.

PR #4552, opened 2021-06-20
Update to Junit5 and allow for mcMMO to be compiled on Java 16

TheBusyBiscuit:
## Premise
Right now mcMMO is not able to be tested successfully when using JDK 16 (tested with AdoptOpenJDK 16 - `16.0.1+9 64-bit 2021-04-20`). Compiling itself works but the unit test `ChunkStoreTest.java` fails.
Here is the full stacktrace if you are interested:
<https://pastebin.com/7DBVLmEZ>

## Proposed change
The reason for this test failure is that it still uses `Junit4` (`junit-vintage-engine` artifact) and `PowerMock` which seem to break on newer Java versions (or at least mcMMO's setup/versions of it).
Updating that Unit Test to use `Junit5` and `Mockito` (which nowadays provides the same capabilities that PowerMock used to provide) fixes the issue entirely.
So while telling people to "use Java 8" to compile is one solution, I reckon a proper update of that unit test to `Junit5` might be a longer-lasting solution in the end. Some other Unit Tests also seem to use `Junit4` but for these the update is more or less just changing the import for `@Test` and removing the `public` modifier from classes and methods.
It would also still work on Java 8 of course.
So as a first draft, I updated everything to JUnit 5 and it seems to work reliably and successfully.

(I initially posted most of these findings on mcMMO's discord but copied them over to here for transparency and reasoning how these changes came about.)

I also posted an issue on the `junit-vintage` bug tracker simultaneously:
https://github.com/junit-team/junit5/issues/2648

### Dependency changes
The following dependencies have changed:

- `maven-surefire-plugin` (`2.22.2` -> `3.0.0-M5`)
- `maven-failsafe-plugin` (`2.22.2` -> `3.0.0-M5`)
- `org.junit.jupiter:junit-jupiter-api` (removed)
- `org.junit.jupiter:junit-jupiter-engine` (removed)
- `org.junit.jupiter:junit-jupiter` (added)
- `org.junit.vintage:junit-vintage-engine` (removed)
- `org.powermock:powermock-module-junit4` (removed)
- `org.powermock:powermock-api-mockito2` (removed)
- `org.mockito:mockito-core` (`3.9.0` -> `3.11.2`)
- `org.mockito:mockito-inline` (added)


TheBusyBiscuit:
I am not quite sure as to why the GitHub action takes this long to complete...
It's nothing to do with the tests at least and it finished much quicker on my fork :/
Perhaps github is just having some performance problems right now.

t00thpick1:
Looks good, didn't realize they replicated power mock into regular mockito.  Update looks like it was pretty straightforward.

PR #4555, opened 2021-06-23
Fix mcmmo picking wrong version on builds withou patch number

### LINKED ISSUE #4554: [BUG] Mcmmo thinks 1.17 is 1.17.1
https://github.com/mcMMO-Dev/mcMMO/blob/de6ba4fb6aa7f04061bbcb6959fe1ef32c82aef3/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java#L61

Hey, nothing serious, for some reason mcmmo thinks when loading that it's running on 1.17.1 and it logs it into the console. I think the problem is in the code above.

image

mcmmo ver - 199
spigot ver - 1.17 - build 3160 (Built using BuildTools, should be latest on 12pm 23rd June)

stepech:
closes #4554 

Probably not even worth PR, but still...
Tested on 1.13, 1.13.2 and 1.17 with correct version printed out.

### DIFF
--- a/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
+++ b/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
@@ -47,7 +47,8 @@ public class PlatformManager {
     private @NotNull MinecraftGameVersion determineGameVersion(String platformVersionString) {
         int major = 0, minor = 0, patch = 0;
 
-        String[] splitVersion = platformVersionString.split("\\.", 3);
+        String[] bukkitAPIVersion = platformVersionString.split("-", 2);
+        String[] splitVersion = bukkitAPIVersion[0].split("\\.", 3);
 
         mcMMO.p.getLogger().info("Platform String: " + platformVersionString);
 
--- a/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
+++ b/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
@@ -48,6 +48,8 @@ public class PlatformManager {
         int major = 0, minor = 0, patch = 0;
 
         mcMMO.p.getLogger().info("Platform String: " + platformVersionString);
+
+        // Gets two numbers separated by . and optional third number after next dot. Must end with - or _
         Matcher versionMatch = Pattern.compile("(\\d+)\\.(\\d+)(?:\\.(\\d+))?[-_].*").matcher(platformVersionString);
 
         if (versionMatch.find()) {
--- a/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
+++ b/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
@@ -56,7 +56,7 @@ public class PlatformManager {
             major = Integer.parseInt(versionMatch.group(1));
             minor = Integer.parseInt(versionMatch.group(2));
 
-            if (!(versionMatch.group(3) == null)) {
+            if (versionMatch.group(3) != null) {
                 patch = Integer.parseInt(versionMatch.group(3));
             }
         }
--- a/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
+++ b/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
@@ -6,9 +6,9 @@ import org.bukkit.Bukkit;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.ArrayList;
 import java.util.Locale;
-import java.util.stream.Collectors;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
Can you perhaps post a list of various version strings you tested this on?

I think this is something that could be unit-tested easily, if we have a sufficient set of testing strings.
I haven't looked at your implementation but it is definitely a good oppurtunity for a unit test, be it here or in a follow-up PR.

TheBusyBiscuit:
Also a little tip: If you use keywords like `closes #4554` or `fixes #4554` etc...
in your pull request's body or commit messages the issue will be automatically closed upon merging.

See here for more details: https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

stepech:
Oh... I am actually totally new to java, so I almost haven't touched the code apart from those commits. What I did was testing compiled versions on dummy spigot servers with versions ▲ as above.
The strings were:
`[13:32:56] [Server thread/INFO]: [mcMMO] Platform String: 1.13.2-R0.1-SNAPSHOT
[13:32:56] [Server thread/INFO]: [mcMMO] Minecraft version determined to be - 1.13.2`
`[13:31:12] [Server thread/INFO]: [mcMMO] Platform String: 1.13-R0.1-SNAPSHOT
[13:31:12] [Server thread/INFO]: [mcMMO] Minecraft version determined to be - 1.13.0`
`[13:26:42] [Server thread/INFO]: [mcMMO] Platform String: 1.17-R0.1-SNAPSHOT
[13:26:42] [Server thread/INFO]: [mcMMO] Minecraft version determined to be - 1.17.0`

(Obviously just the part behind `Platform String:` but just posting these log snippest so that you see how it worked)
That's why I created issue at first.. thought it is hard to solve but this seemed easy enough so I tried...

TheBusyBiscuit:
👍🏻 Perfect, thank you.

These pairs of data should be a good starting point for writing an automated test case
Perhaps I may get bored the next few days, then I can come back and write a unit test for this, thanks.

stepech:
Yeah just carefully, because the whole method is `@Deprecated`.. idk if there is something planned for the future or not..
https://github.com/mcMMO-Dev/mcMMO/blob/d83aaa6a0da21be94da0e6b4e27335a8adb6df93/src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java#L46

TheBusyBiscuit:
Well, a unit test wouldn't need to rely on that method particularly.

The version is read via `Bukkit.getBukkitVersion()`, so what the unit test would do is mock that method, e.g. insert some custom string to be used when that method is called.
Then the `PlatformManager` would do its thing and the unit test compares the resulting `MinecraftGameVersion` of the Platform / PlatformManager to what would be expected for that string.

The implementation itself and that particularly method (which is `private` anyway) is abstracted away here.
And since the `MinecraftGameVersion` is used very widely within mcMMO for version checks, it won't go away anytime soon though. Even if the method itself might be relocated or altered.

That deprecation in particular probably just means that it will be a per-platform version lookup in the future or platform-independent, right now it only works for Bukkit and forks of it. At least that's what I get from looking at the comment.

stepech:
Wait - found much better solution using regex....
try `(\d+)\.(\d+)(?:\.(\d+))*[-_].*` on [regex101](https://regex101.com/)
It would get rid of all the difficult ifs and elses and all the looking and matching all those version numbers. This regex expression puts all into groups so that group1 is major version, group2 is minor version and optional group3 is patch. Also for future reference, group0 is the whole string.
tested with these strings:
1.13.2-R0.1-SNAPSHOT
1.13-R0.1-SNAPSHOT
1.17-R0.1-SNAPSHOT
1.15
1.16-
1.15.2
1.15.2-.....R0.1-SNAPSHOT

Maybe I am too excited about something that won't work at all in the end, but give me some time to rewrite this... It would delete around 60 unnecesary lines and make the whole think easier to read..

stepech:
[Regex Testing](https://regex101.com/r/9N6iJW/3)

src/main/java/com/gmail/nossr50/util/platform/PlatformManager.java
```diff
@@ -47,18 +47,19 @@ public Platform getPlatform() {
...
-        if(splitVersion.length > 2 && Character.isDigit(splitVersion[2].toCharArray()[0]))
-            patch = getSubsequentDigits(splitVersion[2].toCharArray(), 0);
+        // Gets two numbers separated by . and optional third number after next dot. Must end with - or _
+        Matcher versionMatch = Pattern.compile("(\\d+)\\.(\\d+)(?:\\.(\\d+))?[-_].*").matcher(platformVersionString);
+
+        if (versionMatch.find()) {
+            major = Integer.parseInt(versionMatch.group(1));
+            minor = Integer.parseInt(versionMatch.group(2));
+
+            if (!(versionMatch.group(3) == null)) {
```
TheBusyBiscuit, line 59:
```suggestion
            if (versionMatch.group(3) != null) {
```

TheBusyBiscuit (commented):
Using a regular expression for this is definitely the best approach in my opinion.
At first glance the regex looks sufficient to me too, I also wrote a Unit Test to automatically verify the correct detection of Minecraft versions btw, see #4560 

@stepech Thanks for providing some sample strings to test this on :)

stepech:
Hey, I know it's already finished, but just want to throw in and idea here, using `Bukkit,getVersion()` instead of `Bukkit.getBukkitVersion()` we can get strings like git-Spigot-12345-abcdef (MC: 1.17), or some others, which itself is much better for regex, as you can simply check for any numbers behind `MC: ` - I get why this method wasn't used before because of all the gibberish in commit number, build etc. but it might be better for stability and versatility. It's up to you. On my side, I just change regex, which I have prepared already, and put different input into the function, output is the same, but @TheBusyBiscuit you've already done all the checks, so might be much harder for you.
Right now it looks for any two numbers divided by . located before -, with this change it would target directly MC:

TheBusyBiscuit:
Don't worry about the unit test, it's just a draft and can easily be adjusted.

`Bukkit.getVersion()` in combination with a regular expression is also exactly what Paper uses:
https://github.com/PaperMC/PaperLib/blob/master/src/main/java/io/papermc/lib/environments/Environment.java#L45-L46

Whether `Bukkit.getVersion()` or `Bukkit.getBukkitVersion()` is more reliable... might be debatable.
Using `PaperLib` for this might also be an option too but I think `Bukkit.getBukkitVersion()` is fine since I think that's a String which will probably follow the convention more closely than relying on every server software to include `MC: xyz` in their build numbers.

nossr50:
Thank you.

PR #4557, opened 2021-06-24
Add option to not send FakePlayerAnimationEvent for combat skills

tonythemacaroni:
The `FakePlayerAnimationEvent` can cause some compatibility issues with other plugins.


TheBusyBiscuit:
Why not tell the developer of those plugins to ignore fake-events, that's essentially why these are dedicated classes, so they can be ignored easily.

```java
if (mcMMOIsInstalled && event instanceof FakeEvent) {
    return;
}
```

should do the trick and solve any compatibility issues, no need to disable the entire event for that.

nossr50:
Agreed with @TheBusyBiscuit

PR #4560, opened 2021-06-25
Added a Unit Test for mc version detection

TheBusyBiscuit:
## 🧭 Premise
This PR adds an automated parameterized unit test to verify that the Minecraft version is detected correctly.

This test is directly inspired by @stepech's pull request #4555 and should not be merged until that one is.
**As with the current setup of mcMMO, the tests will fail, see issue #4554**.

I think this is a behaviour that should definitely be unit-tested though to protect against future issues like #4554.
The test can also easily be expanded. Though I am not sure what the desired scenario should be for invalid server versions.
Right now it just results in an exception I think, I can also add a test for that if you want, I'd just like to know confirmation that this is the desired behaviour that should happen for when a server version is invalid.

The test itself is pretty simple, it injects a custom platform version string into `Bukkit.getBukkitVersion()`, instantiates a new `PlatformManager` and then compares the determined `MinecraftGameVersion` to the expected result.

These samples were taken directly from the historical data of CraftBukkit's pom.xml file:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/pom.xml
and with the help of @stepech's samples provided in #4555.

Until I checked the history for CB's pom, I also completely forgotten the time when CraftBukkit had `-pre` versions o.O

### 📊 Coverage
While this test seems trivial, it actually has quite a lot of good implications for test coverage.
This single test alone, provides the following additional coverage:

| 📔  Class | 📈  Coverage |
| :-------------------- | :-----: |
| `CompatibilityManager.java` | 67,5% |
| `PlatformManager.java` | 83,0% |
| `PlatformBuilder.java` | 90,0% |
| `BukkitPlatform.java` | 60,0% |
| `MinecraftGameVersion.java` | 52,9% |

### DIFF
--- a/src/test/java/com/gmail/nossr50/util/platform/MinecraftGameVersionTest.java
+++ b/src/test/java/com/gmail/nossr50/util/platform/MinecraftGameVersionTest.java
@@ -1,6 +1,5 @@
 package com.gmail.nossr50.util.platform;
 
-
 import java.util.logging.Logger;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


nossr50:
I really appreciate you taking the time to do this, thank you.

PR #4566, opened 2021-07-11
Update locale_pt_BR.properties

TodFoxBR:
I'm getting an error when trying to run "/mctop Corte de árvore" (I think it's because the whitespace in the skill name).

The word "Lenhador" does not have any whitespace and it has the same meaning.


nossr50:
Thank you

PR #4580, opened 2021-07-22
Set TNT source in blast mining

GriffinCodes:
Using the source in my own plugin; I know the metadata key exists, but I'd rather just use the provided bukkit method

### DIFF
--- a/src/main/java/com/gmail/nossr50/skills/mining/MiningManager.java
+++ b/src/main/java/com/gmail/nossr50/skills/mining/MiningManager.java
@@ -119,6 +119,7 @@ public class MiningManager extends SkillManager {
 
         tnt.setMetadata(mcMMO.tntMetadataKey, mmoPlayer.getPlayerMetadata());
         tnt.setFuseTicks(0);
+        tnt.setSource(player);
         targetBlock.setType(Material.AIR);
 
         mmoPlayer.setAbilityDATS(SuperAbilityType.BLAST_MINING, System.currentTimeMillis());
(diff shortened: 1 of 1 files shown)


lexikiq:
bump! would appreciate this feature as well. would be nice for logging systems, like grief detection or otherwise

lexikiq (approved):

src/main/java/com/gmail/nossr50/skills/mining/MiningManager.java
```diff
@@ -119,6 +119,7 @@ public void remoteDetonation() {
 
         tnt.setMetadata(mcMMO.tntMetadataKey, mmoPlayer.getPlayerMetadata());
         tnt.setFuseTicks(0);
+        tnt.setSource(player);
```
TheBusyBiscuit, line 122:
```suggestion
        if (mcMMO.getCompatibilityManager().getMinecraftGameVersion().isAtLeast(1, 16, 4)) {
            tnt.setSource(player);
        }
```

TheBusyBiscuit (changes requested):
There is a huge issue with this though.

The method `TNTPrimed#setSource()` was added on Oct 25th 2020, source:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/4abf9e96df578e4c842c865e50d0933ce0616938#src/main/java/org/bukkit/entity/TNTPrimed.java

In other words, this method was added in 1.16.3 (-SNAPSHOT, 1.16.4 was the first proper release to have included this) and won't be available in prior Minecraft versions.
That's why I am suggesting a version-gated check here, to make sure there won't be any `NoSuchMethodException`s spit out to servers on earlier versions.

TheBusyBiscuit (approved):

nossr50:
Thank you

PR #4618, opened 2021-08-12
Don't process events with zero final damage

vlaetansky:
I am having a compatibility issue with mcMMO. I pass an event with zero damage and check the continuous events using their name, but instead of the name I expect, I am getting hearts. Not only that, but I can't understand why you limited the zero check only for a specific plugin. My changes should make your plugin more compatible with other plugins that use EntityDamageByEntityEvent.


nossr50:
We want to buff damage when it is reduced to zero, that is the reasoning. Zero is not an indicator of a fake event, you can send out a Fake event (extend or use our FakeEntityDamageByEntityEvent, its an mcMMO class) and mcMMO will ignore it

PR #4619, opened 2021-08-13
Propose: Add an option for Exploiting Fishing.

tunagohan:
## :octocat: Description

When I installed mcMMO, it was very inconvenient that I could not fish more than 10 times in the same place to prevent exploits.
However, if this option is turned off, it is possible to gain an unfairly large amount of experience through fishing exploits.

So I wanted to loosen this restriction, but I couldn't find that option anywhere, so I decided to add it.

At one point, someone playing on my server automatically turned around 90 degrees using mouse macros and such, which anti-exploit measures could not prevent.

So I think that being able to specify the distance you have to move will make the anti-exploit better.

Thank you in advance for your consideration.

I love mcMMO.

## 🔑  Changes

1. The number of times before a warning is displayed can now be specified by the user.
2. The distance that must be traveled can now be specified by the user.

## To Reviewers

Since I don't speak English very well, you may feel uncomfortable with the method names and so on. So if you have any suggestions, please let me know.

## Testing

- [x] Manual Testing
- [x] Build & Package


nossr50:
Thanks

PR #4620, opened 2021-08-15
Trigger change event for party create/disband

PikaMug:
Done for my [latest project](https://github.com/PikaMug/Unite) but useful to developers in general as most party plugins have a create/disband event.


nossr50:
Seems like a good change, thanks.

PR #4630, opened 2021-09-03
Make /party kick case insensitive

Warriorrrr:
Makes the /party kick command case insentive by modifying the hasMember method


Greymagic27 (approved):
Not an expert at Java but this should work fine and should be ready for merge if it's something that's wanted :)

nossr50 (approved):
Thank you

Issue #4516, opened 2021-05-09
[MySQL (MariaDB 8.0)] SQLException: Public Key Retrieval is not allowed


RalphORama:
When using mcMMO with SQL backend and MariaDB >= 8.0, I get the following exception on startup:

```
[20:53:44] [Server thread/ERROR]: [mcMMO] Location: com.mysql.cj.jdbc.exceptions.SQLError createSQLException 110
[20:53:44] [Server thread/ERROR]: [mcMMO] SQLException: Public Key Retrieval is not allowed
[20:53:44] [Server thread/ERROR]: [mcMMO] SQLState: 08001
[20:53:44] [Server thread/ERROR]: [mcMMO] VendorError: 0
```

Other plugins have an option to add `allowPublicKeyRetrieval=true` to the JDBC launch options, but I do not see that option in the mcMMO config. Is there a way to add arbitrary options to the JDBC connection string?

TheBusyBiscuit:
Could you perhaps enable mysql debugging and send the resulting error (if you don't mind)?
`config.yml`:
`MySQL -> Debug: true`

This will give you the full stacktrace.

RalphORama:
```log
[01:41:41 WARN]: java.sql.SQLNonTransientConnectionException: Public Key Retrieval is not allowed
[01:41:41 WARN]:        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:110)
[01:41:41 WARN]:        at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
[01:41:41 WARN]:        at com.mysql.cj.jdbc.ConnectionImpl.createNewIO(ConnectionImpl.java:833)
[01:41:41 WARN]:        at com.mysql.cj.jdbc.ConnectionImpl.<init>(ConnectionImpl.java:453)
[01:41:41 WARN]:        at com.mysql.cj.jdbc.ConnectionImpl.getInstance(ConnectionImpl.java:246)
[01:41:41 WARN]:        at com.mysql.cj.jdbc.NonRegisteringDriver.connect(NonRegisteringDriver.java:198)
[01:41:41 WARN]:        at com.gmail.nossr50.mcmmo.database.tomcat.jdbc.pool.PooledConnection.connectUsingDriver(PooledConnection.java:278)
[...]

PR #4635, opened 2021-09-12
Added support for allowPublicKeyRetrieval=true in the JDBC launch options for MySQL

rosaage:
Added option for allowPublicKeyRetrieval=true as per issue #4516


nossr50:
Seems like a good change, thanks.

PR #4638, opened 2021-09-23
Simplify world blacklist loading

Warriorrrr:
Mainly just a bit of refactoring, this PR makes world blacklist loading a lot simpler/better readable.

Also adds isWorldBlacklisted(String) which isn't used anywhere, but potentially useful for anyone hooking into mcMMO.

### DIFF
--- a/src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
+++ b/src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
@@ -2,29 +2,33 @@ package com.gmail.nossr50.config;
 
 import com.gmail.nossr50.mcMMO;
 import org.bukkit.World;
+import org.jetbrains.annotations.NotNull;
 
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Set;
 
 /**
  * Blacklist certain features in certain worlds
  */
 public class WorldBlacklist {
-    private static final Set<String> blacklist = new HashSet<>();
+    private final Set<String> blacklist = new HashSet<>();
+    private static WorldBlacklist instance;
     private final mcMMO plugin;
+    private final String blackListFileName = "world_blacklist.txt";
 
-    public WorldBlacklist(mcMMO plugin) {
+    public WorldBlacklist(@NotNull mcMMO plugin) {
+        WorldBlacklist.instance = this;
         this.plugin = plugin;
         init();
     }
 
     public void init() {
         //Make the blacklist file if it doesn't exist
-        String blackListFileName = "world_blacklist.txt";
         File blackListFile = new File(plugin.getDataFolder() + File.separator + blackListFileName);
 
         try {
@@ -47,11 +51,11 @@ public class WorldBlacklist {
         }
     }
 
-    public static boolean isWorldBlacklisted(World world) {
-        return blacklist.contains(world.getName().toLowerCase());
+    public static boolean isWorldBlacklisted(@NotNull World world) {
+        return isWorldBlacklisted(world.getName());
     }
 
-    public static boolean isWorldBlacklisted(String worldName) {
-        return blacklist.contains(worldName.toLowerCase());
+    public static boolean isWorldBlacklisted(@NotNull String worldName) {
+        return instance.blacklist.contains(worldName.toLowerCase(Locale.ROOT));
     }
 }
--- a/src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
+++ b/src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
@@ -3,91 +3,53 @@ package com.gmail.nossr50.config;
 import com.gmail.nossr50.mcMMO;
 import org.bukkit.World;
 
-import java.io.*;
-import java.util.ArrayList;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * Blacklist certain features in certain worlds
  */
 public class WorldBlacklist {
-    private static ArrayList<String> blacklist;
+    private static final Set<String> blacklist = new HashSet<>();
     private final mcMMO plugin;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -3,91 +3,55 @@
...
-    private void closeRead(Reader reader) {
-        if(reader != null) {
-            try {
-                reader.close();
-            } catch (IOException e) {
-                e.printStackTrace();
-            }
-        }
+    public static boolean isWorldBlacklisted(World world) {
+        return blacklist.contains(world.getName().toLowerCase());
```
TheBusyBiscuit, line 51: We could simplify this and make use of the other method here

```suggestion
        return isWorldBlacklisted(world.getName());
```

src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -3,91 +3,55 @@
...
-
-        for(String s : blacklist)
-        {
-            if(world.getName().equalsIgnoreCase(s))
-                return true;
-        }
-
-        return false;
+    public static boolean isWorldBlacklisted(String worldName) {
+        return blacklist.contains(worldName.toLowerCase());
```
TheBusyBiscuit, line 55: Missing `Locale` argument
```suggestion
        return blacklist.contains(world.Name.toLowerCase(Locale.ROOT));
```

When doing `.toLowerCase()` or `.toUpperCase()` **always** specify a `Locale`.
Otherwise Java will assume the system-default, which may not be english.

This can lead to unexpected results as with Turkish for example, where `"i".toUpperCase()` will lead to the letter I with a dot over it (`İ`).

Trust me, this will cause problems, especially in projects that are used this widely.

src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -3,91 +3,55 @@
...
         this.plugin = plugin;
-        blacklist = new ArrayList<>();
         init();
     }
 
-    public void init()
-    {
+    public void init() {
         //Make the blacklist file if it doesn't exist
+        String blackListFileName = "world_blacklist.txt";
```
TheBusyBiscuit, line 27: Why are you allocating this String here and not in the class body?

src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -3,91 +3,55 @@
...
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * Blacklist certain features in certain worlds
  */
 public class WorldBlacklist {
-    private static ArrayList<String> blacklist;
+    private static final Set<String> blacklist = new HashSet<>();
```
TheBusyBiscuit, line 17: Why is this even static? It's populated by the `WorldBlacklist` instance.
So to me a Singleton pattern would be more sensible here.

If we want to use it in the static methods below, we could just use the singleton instance like
`instance.blacklist.contains(...)`

src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -3,91 +3,55 @@
...
     private final mcMMO plugin;
 
-    private final String blackListFileName = "world_blacklist.txt";
-
-    public WorldBlacklist(mcMMO plugin)
-    {
+    public WorldBlacklist(mcMMO plugin) {
```
TheBusyBiscuit, line 20: This parameter should probably be annotated as `@NotNull`
```suggestion
    public WorldBlacklist(@NotNull mcMMO plugin) {
```

src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -3,91 +3,55 @@
...
 
-    private void closeRead(Reader reader) {
-        if(reader != null) {
-            try {
-                reader.close();
-            } catch (IOException e) {
-                e.printStackTrace();
-            }
-        }
+    public static boolean isWorldBlacklisted(World world) {
```
TheBusyBiscuit, line 50: This parameter should probably be annotated as `@NotNull`

```suggestion
    public static boolean isWorldBlacklisted(@NotNull World world) {
```

src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -3,91 +3,55 @@
...
-    {
-
-        for(String s : blacklist)
-        {
-            if(world.getName().equalsIgnoreCase(s))
-                return true;
-        }
-
-        return false;
+    public static boolean isWorldBlacklisted(String worldName) {
```
TheBusyBiscuit, line 54: This parameter should probably be annotated as `@NotNull`

```suggestion
    public static boolean isWorldBlacklisted(@NotNull String worldName) {
```

TheBusyBiscuit (changes requested):
Some nitpickings.

I highly urge you to look into that `Locale` issue though.
This can lead to a lot of unexpected results and issues down the line if it gets overlooked.

If you need any documents outlining this particular issue, I'd recommend you to give this a read: https://wiki.sei.cmu.edu/confluence/display/java/STR02-J.+Specify+an+appropriate+locale+when+comparing+locale-dependent+data

src/main/java/com/gmail/nossr50/config/WorldBlacklist.java
```diff
@@ -2,92 +2,60 @@
...
-                if(currentLine.length() == 0)
-                    continue;
-
-                if(!blacklist.contains(currentLine))
-                    blacklist.add(currentLine);
-            }
-
+            while((currentLine = reader.readLine()) != null)
+                if (!currentLine.isEmpty())
+                    blacklist.add(currentLine.toLowerCase());
```
TheBusyBiscuit, line 47:
```suggestion
                    blacklist.add(currentLine.toLowerCase(Locale.ROOT));
```

TheBusyBiscuit (changes requested):
Thanks for addressing the concerns I mentioned, seems like you missed one line though

PR #4647, opened 2021-11-05
Add deprecated constructor w/o startinglevel per d9e195f

PikaMug:
Not sure why this was made a TODO as opposed to just doing the thing, but here we go. This will give resources which depend on mcMMO more time to update to the new constructors.


nossr50:
Thanks

Issue #4105, opened 2020-01-16
Fishing shake percentage seems off.


andrewkm:
Using 2.1.113 compiled on https://github.com/mcMMO-Dev/mcMMO/commit/406429f4e3cbe318f9c593b714b308408b6ffd1e

At max rank, chance is 56%
https://i.imgur.com/zItFapn.png
Where is this 56% number even coming from... am I missing something?

Advanced.yml
```
    Fishing:

        ShakeChance:
            Rank_1: 15.0
            Rank_2: 15.0
            Rank_3: 25.0
            Rank_4: 35.0
            Rank_5: 45.0
            Rank_6: 55.0
            Rank_7: 65.0
            Rank_8: 75.0
```

Wiki states 75% as well:
https://mcmmo.fandom.com/wiki/Fishing

andrewkm:
In depth test:
Default paper, default mcMMO and not even a permissions plugin, straight up following mcmmo's plugin.yml

```
<fishing level>:<shake chance>
124:0%
125:0%
250:2%
350:6%
400:6%
800:30%
1000:56%
5000:56%
1000000:56%
```

https://i.imgur.com/JtoEc97.png
Conclusion: It's definitely an mcMMO bug.

nossr50:
There's some unused settings in the config, with that said, I'll investigate and see what is going on.

andrewkm:
Any updates on this one @nossr50

Lyther:
Spotted this problem in the source code.

The fishing skill uses **RandomChanceStatic** to calculate the chances. However, this class has a mistaken definition with `probabilityCap = XPos`. After calculation in **getChanceOfSuccess**, the value `probabilityCap` is put into `maxProbability`. Leads to result formula `chance = XPos * (XPos / 100.0D)`.

This explains why the bug gives chance `definition ^ 2 / 100` as a result.

Fixed this problem in my pull request.

andrewkm:
@Lyther @nossr50 
The above commit did not fix the issue.
Instead, fishing shake is now ALWAYS stuck to 30% regardless of your level.
It unlocks at level 250 and simply never increases, whether your fishing level 1000, level 10,000, etc.

The above PR broke it even more.

Lyther:
[...]

PR #4649, opened 2021-11-08
Fix issue #4105 Fishing shake percentage seems off.

Lyther:
The fishing skill uses `RandomChanceStatic` to calculate the chances. **However, this class has a mistaken definition with probabilityCap = XPos.** After calculation in `getChanceOfSuccess`, the value `probabilityCap` is put into `maxProbability`. Leads to result formula `chance = XPos * (XPos / 100.0D)`.

This explains why the bug gives chance `definition ^ 2 / 100 as a result`.

The fixed code now can work properly.


nossr50:
Thank you for taking time to fix this

PR #4650, opened 2021-11-08
Fix brew event error, update to 1.17.1 API

gecko10000:


gecko10000:
Fixed display names not formatting properly

nossr50:
> Fixed display names not formatting properly

Would you mind submitting this in its own PR?

Also thanks for submitting a PR for the alchemy issues, but your changes would break older versions. I'll do some work on the issue myself.

PR #4651, opened 2021-11-09
Fix hex colors in broadcasts

gecko10000:


nossr50:
Thanks gecko, if you submit more PRs please add a detailed description

PR #4655, opened 2021-11-10
Fix alchemy not working with latest changes

### LINKED ISSUE #4653: Alchemy skill not working with version 2.1.203
After updating MCMMO to version 2.1.203 I realized that the Alchemy skill was not working anymore. 
Brewing potions would give me the following error.
[error.log](https://github.com/mcMMO-Dev/mcMMO/files/7509433/error.log)

After taking a look a the file AlchemyPotionBrewer.java I saw that something was changed in commit 8eee39b88b95a0bb8fddc0d0ae90ae9deeededfd which is probably breaking this feature.

We are creating an ArrayList at line 117, but we never define his size nor add objects to it. We later access this array at line 132 but since we've not changed the size of it we get an access out of bound.

Changing the line 117 to something similar to 
`var outputList = new ArrayList<ItemStack>(3);` will probably fix the issue.

Thank you :)

gecko10000:
This PR modifies the `AlchemyPotionBrewer.finishBrewing` method to add the output item to the output list instead of setting the item to the index (which errors and causes an `IndexOutOfBoundsException` as the list doesn't have the index to set). This fixes #4653 and works properly on my testing server.

### DIFF
--- a/src/main/java/com/gmail/nossr50/skills/alchemy/AlchemyPotionBrewer.java
+++ b/src/main/java/com/gmail/nossr50/skills/alchemy/AlchemyPotionBrewer.java
@@ -129,7 +129,7 @@ public final class AlchemyPotionBrewer {
             inputList.add(input);
 
             if (output != null) {
-                outputList.set(i, output.toItemStack(item.getAmount()).clone());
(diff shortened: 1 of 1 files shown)


nossr50:
Thank you

PR #4656, opened 2021-11-10
Added 1.13 to 1.17 mob loot to shake

dodieboy:
This PR added support for 1.13 to 1.17 mob to fishing skill shake. I did not add Ravager as it only drop saddle and i think it may be too broken if a player can easily get saddle with it, how do you all think?

### DIFF
--- a/src/main/resources/fishing_treasures.yml
+++ b/src/main/resources/fishing_treasures.yml
@@ -1,6 +1,6 @@
 #
 #  Settings for Fishing Treasures / Shake Treasures
-#  Last updated on 12/28/2020
+#  Last updated on 11/10/2021
 ###
 Fishing:
     LEATHER_BOOTS:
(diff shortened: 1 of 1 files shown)


Axyzel:
Not sure ravager would be that broken, especially if there's a way to only get one from it. You can already get copious amounts of saddles from normal fishing.

dodieboy:
from what i see how the shake work, if the mob only have 1 item drop, it will get the item 100% of the time when shake activate. So a player can get like about 40 saddle if you are lucky on max fishing level

dodieboy:
Did some test if my shake is 30% i can get 11 saddle from 1 ravager

dodieboy:
Tried to add tipped arrow to stray but the potion effect is not applying to the item, did I did anything wrong?

image

Axyzel:
Point taken on the ravager. Saddles aren't mega useful but that is kinda ridiculous, at high level that's more powerful than a raid farm.
On the note of the stray, you may be using the wrong metadata tags for the kind of potion? It might just be Potion not PotionType. Not certain but that's probably where the issue is?

dodieboy:
If mcmmo is using the naming from spigot I believe the effect should be "SLOWNESS" and I believe it is in the "PotionData" base on the spigot api and the potion for witch in this config file is also using "PotionData"

https://hub.spigotmc.org/javadocs/spigot/org/bukkit/potion/PotionType.html

gecko10000:
Compatibility for all pre-1.17.1 versions was removed with the most recent update, so I don't see how this will help with anything.

dodieboy:
All mob from pre1.17.1 will still be in 1.17.1, how does this PR not help in anything?

gecko10000:
Sorry, I thought you meant you were adding support for each version, not the new mobs. My bad.

PR #4658, opened 2021-11-10
Fixed the alchemy properly

NemuruYama:
The fix made by the other person missed an edge case for if someone only was brewing 1 or 2 potions, with the potions not being from left to right.

Instead of just adding the potions, I am now filling the arraylist with empty items and then doing a null check later.

I am not familiar with programming MC plugins, so if there is a better way to fill the array with empty items, be sure to change it.

With kind regards,
Lucas


dodieboy:
tested that it fix #4660

nossr50 (commented):
With a few edits this looks good

nossr50 (approved):

Issue #4667, opened 2021-11-14
Herbalism Issue (Lilypad + Beetroot)


Flibknort:
Beetroot isn't affected by Double Drops but still give xp and auto replant.
Lilypads on the other hand, doesn't give xp and doesn't work with herbalism in general. No Double drop, no xp, no interaction. 
Diagnosing via `/mmodebug`, Lilypads still give "plant processed : 1" but after that nothing else shows up.
![image](https://user-images.githubusercontent.com/94317388/141695017-fe06bc11-044c-423f-9c64-a1806ddf1ca5.png)

Running on Airplane-92 (R0.1 Snapshot) 1.17.1, mcMMO v204
![image](https://user-images.githubusercontent.com/94317388/141695094-0f791165-9742-4a6f-a36a-f95495f6991e.png)

Lyther:
Working on this issue.

By the definition of `Material.class`, BEETROOT is an ageable plant, while LILY_PAD isn't:

```java
BEETROOTS(22075, Ageable.class),
LILY_PAD(19271),
```

The lily_pad only shows up in the debug mode `plants processed`, but never actually processed as a plant.

I'll fix the beetroot's double drop properly, but not going to add lily_pad as a plant.

Chew:
Fixed in #4669 I suppose?

PR #4669, opened 2021-11-16
fix issue #4667 Herbalism Issue (Beetroot)

Lyther:
## General Problem

I noticed that in the definition of `Material.class`, there's difference between `BEETROOT` and `BEETROOTS`. `BEETROOT` is the item that a player holds in hand, while `BEETROOTS` is the plant block that grows.

My fix involves these 2 files:

### src/main/java/com/gmail/nossr50/commands/skills/HerbalismCommand.java

According to the `config.yml` and the `plugin.yml`, the Material used in the `Green_Thumb_Replanting_Crops` check should be `BEETROOTS` instead of `BEETROOT`. I fixed the judgment from `BEETROOT` to `BEETROOTS`.

### src/main/java/com/gmail/nossr50/util/ItemUtils.java

Here I fixed issue #4667 Herbalism Issue (Lilypad + Beetroot). The reason why beetroot hasn't given double drops is that the `isHerbalismDrop` checks `BEETROOT` instead of `BEETROOTS`. I added `BEETROOTS` to the list.

Now in my test, the fix works properly.

Hope you make a full recovery from the cold.

### DIFF
--- a/src/main/java/com/gmail/nossr50/commands/skills/HerbalismCommand.java
+++ b/src/main/java/com/gmail/nossr50/commands/skills/HerbalismCommand.java
@@ -91,7 +91,7 @@ public class HerbalismCommand extends SkillCommand {
     protected void permissionsCheck(Player player) {
         hasHylianLuck = canUseSubskill(player, SubSkillType.HERBALISM_HYLIAN_LUCK);
         canGreenTerra = Permissions.greenTerra(player);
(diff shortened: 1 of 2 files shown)


nossr50:
Thanks, I'm a bit confused on the title, did you mean to also remove my lily pad exploit fix? It's not in your changes. I'm still getting over the cold 😷

Lyther:
Not changing the Lily_pad. Only works on the beetroot.

PR #4676, opened 2021-12-04
lithuanian translation for mcmmo

dexasz:
first time making a PR, if something's wrong with the translation feel free to message me. I'll fix it up.


nossr50 (approved):
Thanks, we will have to see if unicode is natively supported or else I might have to UTF-16 this

PR #4693, opened 2021-12-21
Added option to disallow using enchanted materials to repair items.

JeBobs:
Due to an issue where special use items created by other plugins were being used to repair tools, I created this option to allow server owners to disable using enchanted items to repair tools.

Note, this does not disable repairing enchanted _tools_, just using enchanted _materials_ to repair tools.
 
Since to my knowledge you can not (through vanilla survival means) enchant materials (i.e. iron ingots) used to repair tools, having this disabled should not affect normal gameplay.

Here is what I have done in each affected file:
AdvancedConfig  - Added getAllowEnchantedRepairMaterials to config, enable it if you'd like to mimic the original behavior
RepairManager  - Added functionality to (if enabled) disallow enchanted materials to be used to repair tools, and find potentially non-enchanted repair materials (if they exist in the player's inventory)

Feel free to let me know if there is anything you would like changed and I will be happy to do so. 😃

### DIFF
--- a/src/main/java/com/gmail/nossr50/skills/repair/RepairManager.java
+++ b/src/main/java/com/gmail/nossr50/skills/repair/RepairManager.java
@@ -28,8 +28,11 @@ import org.bukkit.entity.Player;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
--- a/src/main/java/com/gmail/nossr50/skills/repair/RepairManager.java
+++ b/src/main/java/com/gmail/nossr50/skills/repair/RepairManager.java
@@ -28,11 +28,8 @@ import org.bukkit.entity.Player;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
 
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Objects;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/gmail/nossr50/skills/repair/RepairManager.java
```diff
@@ -128,6 +133,34 @@ public void handleRepair(ItemStack item) {
...
+            if (toRemove.getEnchantments().size() > 0) {
+
+                // Lots of array sorting to gather a possible list of items not including the enchanted item.
+                List<ItemStack> possibleMaterials = Arrays.stream(inventory.getContents())
+                        .filter(Objects::nonNull)
+                        .filter(p -> p.getType() == repairMaterial)
+                        .filter(p -> p.getEnchantments().size() == 0).toList();
+
+                // Fail out with "you need material" if we don't find a suitable alternative.
+                if (possibleMaterials.size() == 0) {
```
TheBusyBiscuit, line 151: A general tip:
`x.isEmpty()` and `!x.isEmpty()` are generally better suitable instead of `x.size() == 0` or `x.size() > 0`.
They are simply more descriptive of the actual check you want to perform.

src/main/java/com/gmail/nossr50/skills/repair/RepairManager.java
```diff
@@ -128,6 +133,34 @@ public void handleRepair(ItemStack item) {
...
+        if (!mcMMO.p.getAdvancedConfig().getAllowEnchantedRepairMaterials()) {
+
+            // See if our proposed item is even enchanted in the first place.
+            if (toRemove.getEnchantments().size() > 0) {
+
+                // Lots of array sorting to gather a possible list of items not including the enchanted item.
+                List<ItemStack> possibleMaterials = Arrays.stream(inventory.getContents())
+                        .filter(Objects::nonNull)
+                        .filter(p -> p.getType() == repairMaterial)
+                        .filter(p -> p.getEnchantments().size() == 0).toList();
```
TheBusyBiscuit, line 148: I don't see a reason to make a `List<ItemStack>` here, if the list isn't empty, you only ever access the first element.
Therefore `Stream#findFirst()` makes much more sense here, it returns an `Optional`, allowing you to still check whether a candidate was found or not.

```suggestion
                Optional<ItemStack> possibleMaterial = Arrays.stream(inventory.getContents())
                        .filter(Objects::nonNull)
                        .filter(p -> p.getType() == repairMaterial)
                        .filter(p -> p.getEnchantments().isEmpty())
                        .findFirst();
```

(The `isEmpty` filter can probably be passed as a method reference too but just as a proof of concept)

TheBusyBiscuit (changes requested):

nossr50:
Thank you

PR #4703, opened 2022-01-03
Add Multiple Custom Boost Options in Experience.yml

TreemanKing:
Replaced single custom boost with configurable multiple options. This can be added by adding the name of the rank and double to the configuration in the required spot.


TreemanKing:
Whilst thinking about this in detail, this kind of makes all the preset boosts (quadruple, double) redundant. Any thoughts?

nossr50:
I appreciate the effort but a better solution was written by me in the configurable branch, it's practically copy paste-able too if you want to do the legwork of porting it to master

TreemanKing:
Will do so now! Thanks!

PR #4707, opened 2022-01-05
add and translate missing strings to Lithuanian language (also fix a typo that broke acrobatics placeholders)

dexasz:
I didn't notice at first. Apparently some strings were missing in Lithuanian translation files, so I added and translated them.

Also fixed a typo that broke acrobatics placeholders, due to me adding an extra "S" accidentally lol


nossr50:
Thank you

PR #4713, opened 2022-01-07
Disable parsing comments to prevent duplicate comments

### LINKED ISSUE #4702: YAML config issues (too many comments)
So, recently some people seemed to start having issues with mcMMO config files not being able to load properly.

https://pastebin.com/CBM4Hf0M

I investigated the issue on discord and found out that their config files had a lot of duplicated comments.

<details>
<summary>Example experience.yml (as posted by `@Ironic_8b49#9887` on discord)</summary>

```yaml
ExploitFix:
    # Prevent many exploits related to fishing
    # Prevent many exploits related to fishing
    # Prevent many exploits related to fishing
    # Prevent many exploits related to fishing
    # Prevent many exploits related to fishing
    Fishing: true
    EndermanEndermiteFarms: true
    Acrobatics: true
    LavaStoneAndCobbleFarming: true
    PistonCheating: true
    UnsafeEnchantments: false
    COTWBreeding: true
    TreeFellerReducedXP: true
    SnowGolemExcavation: true
    # This include NPCs from stuff like Citizens, this is not a setting for Vanilla Minecraft Villagers (Which can be considered NPCs)
    # mcMMO normally doesn't process attacks against an Entity if it is an NPC from another plugin
    # Of course, mcMMO doesn't know for sure whether or not something is an NPC, it checks a few known things, see our source code to see how
    # This include NPCs from stuff like Citizens, this is not a setting for Vanilla Minecraft Villagers (Which can be considered NPCs)
    # mcMMO normally doesn't process attacks against an Entity if it is an NPC from another plugin
    # Of course, mcMMO doesn't know for sure whether or not something is an NPC, it checks a few known things, see our source code to see how
    # This include NPCs from stuff like Citizens, this is not a setting for Vanilla Minecraft Villagers (Which can be considered NPCs)
    # mcMMO normally doesn't process attacks against an Entity if it is an NPC from another plugin
    # Of course, mcMMO doesn't know for sure whether or not something is an NPC, it checks a few known things, see our source code to see how
    # This include NPCs from stuff like Citizens, this is not a setting for Vanilla Minecraft Villagers (Which can be considered NPCs)
    # mcMMO normally doesn't process attacks against an Entity if it is an NPC from another plugin
    # Of course, mcMMO doesn't know for sure whether or not something is an NPC, it checks a few known things, see our source code to see how
    # This include NPCs from stuff like Citizens, this is not a setting for Vanilla Minecraft Villagers (Which can be considered NPCs)
    # mcMMO normally doesn't process attacks against an Entity if it is an NPC from another plugin
    # Of course, mcMMO doesn't know for sure whether or not something is an NPC, it checks a few known things, see our source code to see how
    PreventPluginNPCInteraction: true
```

</details>

As seen in the above snippet, some comment lines were duplicated a couple of times.
I looked at the source code of snakeyaml and it appears that any config node with more than 100 comments will cause the internal event queue to fill up with CommentEvents. And since snakeyaml's internal queue is of size 100, an overload will cause this error to occur. This results in the above exception "queue is full" to be thrown when serializing the config.

(src: https://bitbucket.org/snakeyaml/snakeyaml/src/2e2c75da65cc5d15662fdca03e465a6148847182/src/main/java/org/yaml/snakeyaml/emitter/Emitter.java#lines-190)

In other words, yaml config files break when there are more than ~100 successive comment lines with no data inbetween.
Certainly not mcMMO's fault but a very odd error for sure.

For anyone interested, I set up a minimal demo of this issue with nothing but snakeyaml and were able to reproduce it:
https://gist.github.com/TheBusyBiscuit/1cc621aff4e0f39d642abd5703dd96db

This demonstrates that this is not an issue within Paper, Spigot or mcMMO but a limitation/issue from snakeyaml itself.
However I reckon that mcMMO could circumvent this by preventing these comments from being duplicated.
I have yet to find the cause for this but I wanted to share my findings here on github, so we can investigate this issue further.
Perhaps we will be able to push a fix before too many server owners are affected by this, as this is a *breaking* issue in the sense that it can stop mcMMO from loading up.

Chew:
This apparently also removes all duplicate comments that were added.

So far tests:

* 1 - Works
* 0 - Not works

Would love more testers, though, since I can't reproduce this issue.

Fixes #4702

### DIFF
--- a/src/main/java/com/gmail/nossr50/config/AutoUpdateConfigLoader.java
+++ b/src/main/java/com/gmail/nossr50/config/AutoUpdateConfigLoader.java
@@ -48,6 +48,12 @@ public abstract class AutoUpdateConfigLoader extends ConfigLoader {
         super.loadFile();
         FileConfiguration internalConfig = YamlConfiguration.loadConfiguration(mcMMO.p.getResourceAsReader(fileName));
 
+        try {
+            internalConfig.options().parseComments(false);
+            config.options().parseComments(false);
+        }
+        catch (NoSuchMethodError ignored) {}
+
         Set<String> configKeys = config.getKeys(true);
         Set<String> internalConfigKeys = internalConfig.getKeys(true);
 


the456gamer:
I will probably open my open pr, as I have a different way to fix this

Reproduction needs: a config file with a user added key:
Like adding ```testing: true```
To the main config.yml will duplicate comments on later loads

the456gamer:
> This apparently also removes all duplicate comments that were added.

Doesn't this mean newly generated config files (on first run) would also be missing the comments?
Or is the change backwards compatible

TheBusyBiscuit (approved):
quick and easy fix

the456gamer:
Not a code review, but functionality: on config file save, this will wipe out all custom comments
This bad, esp if the uses for example comments out a section of a config file

I haven't tested the original on an older server, so this could be old behaviour before the comment duplication, but it is not preferred imo

the456gamer:
> > This apparently also removes all duplicate comments that were added.
> 
> Doesn't this mean newly generated config files (on first run) would also be missing the comments? Or is the change backwards compatible

Also disregard this as I thought the wiping out means that it would rely on something else, not force resetting

Chew:
This should fully mimic the old behavior, as it's simply turning off the new one. Would love more testing on this.

PR #4763, opened 2022-04-10
Fixes Java 17 compiling on ovh server

jrodr10:
Maven compiler plugin was outdated for java 17 compilation, no idea why it worked on the actions but not my local server but regardless this resolves that. tested and runs no error


nossr50:
I think I misunderstood what this PR was about in discord, sorry about that.

PR #4766, opened 2022-04-22
Fix hanging plants searching up instead of down.

### LINKED ISSUE #4765: Herbalism XP Exploit
Minecraft Version: 1.18.2
Purpur-1609
mcMMO version: 2.1.211
No other plugins, default mcMMO config

Breaking Weeping Vines seems to give Herbalism XP for all the Weeping Vines above them. Essentially what happens is that when you break the bottommost full plant - that is, excluding the final, tip part -, you get Herbalism XP for all the sections above it. So while you are only breaking 1 block, you can get XP for hundreds that you aren't actually breaking.

This exploit allows players to create automated farms that allow them to max out the Herbalism in a matter of minutes, just standing still and holding M1.

Here is a video demonstration: https://youtu.be/ulGWGtJZtB8

Warriorrrr:
Closes #4765

Fixes hanging plants rewarding more xp than intended.


nossr50:
Looks good, thank you!

Issue #4379, opened 2021-01-11
[Request] mcmmo.perks.xp.25percentboost


WhatsTheBadNews:
As the title states, I would very much appreciate it if `mcmmo.perks.xp.25percentboost` was added. 😃

Thank you for all your had work nossr!

PR #4773, opened 2022-06-01
Added 25% xp boost permission

smuddgge:
Added perk for `[Request] mcmmo.perks.xp.25percentboost #4379`

- Added 25% xp boost in `plugin.yml`
- Added permission method `oneAndAQuarterXp`
- Added a check in PerksUtils for the permission


nossr50:
Looks good

Issue #4675, opened 2021-12-04
/mctop can be used to know vanished staff are online


WhatsTheBadNews:
![image](https://user-images.githubusercontent.com/29821149/144698703-b78af00a-f4af-44ab-afc8-505ae91d306a.png)
^ in the screenshot above, the Mod Bokuuto is vanished. But a regular player can use /mctop to know they are vanished. The prefix and color shows they are online.

Possible solutions:
1.) Don't show the color/prefix for vanished players in /mctop
2.) Add a permission we can give to mods to prevent them from showing up on /mctop

Relevant Versions:
MC 1.17.1
Paper 388
mcMMO 2.1.205
PremiumVanish 2.7.7

PR #4774, opened 2022-06-01
Fixed scoreboard showing vanish players bug

KingCreator11:
This is a bit of a hack to fix the following issue: https://github.com/mcMMO-Dev/mcMMO/issues/4675. It turns out that issue was caused by another plugin (potentially very frequently used) overloading the Scoreboard `getScore` [method](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/scoreboard/Objective.html#getScore(java.lang.String)). This PR only has a single commit which adds an else statement to add a space + a white color code in front of the mctop scoreboard.

I completely understand if you choose to deny this PR as it isn't really an MCMMO issue but I do believe it can help others with the same issue.

### DIFF
--- a/src/main/java/com/gmail/nossr50/util/scoreboards/ScoreboardWrapper.java
+++ b/src/main/java/com/gmail/nossr50/util/scoreboards/ScoreboardWrapper.java
@@ -20,7 +20,6 @@ import com.gmail.nossr50.util.skills.SkillTools;
 import org.apache.commons.lang.Validate;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
-import org.bukkit.OfflinePlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.scheduler.BukkitRunnable;
 import org.bukkit.scheduler.BukkitTask;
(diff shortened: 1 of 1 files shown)


src/main/java/com/gmail/nossr50/util/scoreboards/ScoreboardWrapper.java
```diff
@@ -20,6 +20,7 @@
 import org.apache.commons.lang.Validate;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
+import org.bukkit.OfflinePlayer;
```
nossr50, line 23: Can you remove this unused import?

nossr50 (changes requested):
Looks good, do you mind making the changes requested?

src/main/java/com/gmail/nossr50/util/scoreboards/ScoreboardWrapper.java
KingCreator11: Removed it!

Greymagic27 (approved):

nossr50:
I have to review the implications of this change further before merging

nossr50:
I don't think I can accept this PR as I think it could have some unintended side effects.. I wish I could though, if you put in a plugin check (to see if that other plugin is running) then I could add this safely

PR #4779, opened 2022-06-09
Fix Maven build failure

abridgedcarp:
This fixes the Maven build failure. It also appears to get rid of some console errors I was experiencing (i.e., `[Thu 12:14:00 ERROR Minecraft] Could not pass event PlayerInteractEvent to mcMMO v2.1.213-SNAPSHOT`) while running the mod on a 1.19 server


nossr50:
Thank you

PR #4790, opened 2022-06-27
Added MMODebug permission

smuddgge:
- Added the `mcmmo.commands.mmodebug` for the command `/mmodebug`

Seems to default to true on my test server. Therefor, it provides an option for admins to disable it.


nossr50:
This command has no permission requirements on purpose, makes my life easier. I'm willing to listen to a good argument for having a permission added... but I lean towards keeping it how it is

PR #4791, opened 2022-07-02
Switch to new sk89q repo

PikaMug:
Seems like https://maven.sk89q.com/repo/ is down for good, causing projects which depend on mcMMO to fail build with "Cannot resolve com.sk89q.worldedit.worldedit-libs:core:7.2.0-SNAPSHOT"


TheBusyBiscuit (approved):

nossr50:
Thank you

PR #4797, opened 2022-07-17
Updated kyori adventure-platform dependencies

### LINKED ISSUE #4792: Hovering over skills in chat output does not return anything.
Paper dev 56 (1.19)
McMMO dev 1313 (https://github.com/mcMMO-Dev/mcMMO/commit/2c09a0fdb64d5df0032e950bbd20c5fe2d1ca96d)

You use to be able to hover over @ sub-skills and info in various mcmmo message output such as ``/mining`` -> ``@Double Drops``. However as of 1.19, you can no longer hover over things. 

Adventure library dependency needs to be updated and compiled against? ShowItem plugin seemingly fixed this problem by doing so.
> Fixed that the items in commands weren't clickable and had no hover by updating the adventure library and platform to the latest version.

Greymagic27:
An attempt to fix #4792 and a version update nonetheless. 

Taken new versions from: https://search.maven.org/search?q=g:net.kyori%20AND%20a:adventure-platform*

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -279,12 +279,12 @@
         <dependency>
             <groupId>net.kyori</groupId>
             <artifactId>adventure-platform-bukkit</artifactId>
-            <version>4.0.1</version>
+            <version>4.1.1</version>
         </dependency>
         <dependency>
             <groupId>net.kyori</groupId>
             <artifactId>adventure-platform-api</artifactId>
-            <version>4.0.1</version>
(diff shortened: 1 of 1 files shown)


nossr50:
I'm not sure if this will fix #4792 so let me know if it does

Greymagic27:
Will do. I'll load up my test server and check if it's worked!

PR #4826, opened 2022-09-14
Lumber bonus blocks

destro174:
Apply item in mainhand to lumber bonus blocks. This allows item enchants to be applied to these blocks,


nossr50:
Thank you

PR #4833, opened 2022-10-18
Fix bug where players could know staff are invis

DarkKnights22:
When inspecting a staff member that is invisible to you, the displayed message is different from when inspecting an actual offline player; the power level is shown and the heading other message does not mention the player is offline whereas the offline message specified the player is offline.

### DIFF
--- a/src/main/java/com/gmail/nossr50/commands/player/InspectCommand.java
+++ b/src/main/java/com/gmail/nossr50/commands/player/InspectCommand.java
@@ -74,7 +74,7 @@ public class InspectCommand implements TabExecutor {
                 if (!isVanished && CommandUtils.tooFar(sender, target, Permissions.inspectFar(sender))) {
                     return true;
                 }
-                
+
                 if (mcMMO.p.getGeneralConfig().getScoreboardsEnabled()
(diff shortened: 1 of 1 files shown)


src/main/java/com/gmail/nossr50/commands/player/InspectCommand.java
```diff
@@ -74,7 +74,7 @@ public boolean onCommand(@NotNull CommandSender sender, @NotNull Command command
                 if (!isVanished && CommandUtils.tooFar(sender, target, Permissions.inspectFar(sender))) {
                     return true;
                 }
-
+
```
Greymagic27, line 77: Why the extra dead space here? One return works to ensure it indents.

DarkKnights22: Whoops, I had whitespace changes hidden. Fixed.

Greymagic27: Hehe ty

Greymagic27 (approved):

nossr50:
Thank you

PR #4835, opened 2022-10-23
Add ability to change party chat colour

DarkKnights22:
Added the ability to change the color of messages sent in the party chat system.


nossr50:
Don't we do this in locale? I could be wrong.

PR #4839, opened 2022-11-29
Dupe Fix: Don't add bonus drops for non-natural blocks

Elioby:
Currently, the double drops sub-skill gives you double drops regardless of if the block is naturally generated or placed by a player. This PR fixes that by doing a check before dropping bonus items to ensure the block was not placed by a player.


Elioby:
Turns out there must be another issue going on here, the place store check is already done before the block is marked as with bonus drops.

PR #4840, opened 2022-12-05
Endgame Update

nossr50:
A while back I started working on an "endgame" update for mcMMO, I will be moving this into master as 2.2.000 beta soon, the idea of endgame is to provide new skill progression after level 1000 (default soft cap), only a few skills have an endgame mechanic, but more will be added when I find time, I will merge into master after fixing compiling errors + manual testing + unit tests.


nossr50:
Near future todo: Fix command probability display on the new mastery skills, double check counter attack RNG, write unit tests for RNG mechanics

src/main/resources/advanced.yml
```diff
@@ -590,9 +590,16 @@ Skills:
             Knock_On_Wood:
                 Add_XP_Orbs_To_Drops: true
 
+        # Triple Drops
+        CleanCuts:
+            # ChanceMax: Maximum chance of receiving double drops (100 = 100%)
+            # MaxBonusLevel: Level when the maximum chance of receiving double drops is reached
```
dod0lp, line 610: this should say triple* drops

Issue #4767, opened 2022-05-03
Invalid EntityDamageByEntityEvent damager being set by McMMO


HexedHero:
```
[12:45:52 ERROR]: Could not pass event FakeEntityDamageByEntityEvent to CursedCore v1.6.5
java.lang.NullPointerException: Cannot invoke "org.bukkit.entity.Entity.getType()" because "damager" is null
        at dev.hexedhero.cursedcore.listeners.additions.CombatTag.a(CombatTag.java:144) ~[CursedCore-1.6.5.jar:?]
        at com.destroystokyo.paper.event.executor.asm.generated.GeneratedEventExecutor41.execute(Unknown Source) ~[?:?]
        at org.bukkit.plugin.EventExecutor.lambda$create$1(EventExecutor.java:75) ~[paper-api-1.18.2-R0.1-SNAPSHOT.jar:?]
        at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:80) ~[paper-api-1.18.2-R0.1-SNAPSHOT.jar:git-Paper-312]
        at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:70) ~[paper-api-1.18.2-R0.1-SNAPSHOT.jar:?]
        at org.bukkit.plugin.SimplePluginManager.callEvent(SimplePluginManager.java:669) ~[paper-api-1.18.2-R0.1-SNAPSHOT.jar:?]
        at com.gmail.nossr50.util.skills.CombatUtils.canDamage(CombatUtils.java:887) ~[mcMMO.jar:?]
        at com.gmail.nossr50.util.skills.CombatUtils.dealDamage(CombatUtils.java:576) ~[mcMMO.jar:?]
        at com.gmail.nossr50.util.skills.CombatUtils.dealDamage(CombatUtils.java:532) ~[mcMMO.jar:?]
        at com.gmail.nossr50.skills.woodcutting.WoodcuttingManager.processTreeFeller(WoodcuttingManager.java:125) ~[mcMMO.jar:?]
        at com.gmail.nossr50.listeners.BlockListener.onBlockBreak(BlockListener.java:401) ~[mcMMO.jar:?]
        at com.destroystokyo.paper.event.executor.asm.generated.GeneratedEventExecutor676.execute(Unknown Source) ~[?:?]
[...]

PR #4863, opened 2023-02-09
Don't call EntityDamageByEntityEvent with null damager

Elioby:
Currently, canDamage is called with the nullable argument attacker in the following method:
https://github.com/mcMMO-Dev/mcMMO/blob/a58c6be8a09b29a9c2a338e3dff4d208db7e6135/src/main/java/com/gmail/nossr50/util/skills/CombatUtils.java#L572-L582

This causes other plugins to throw NPEs when trying to get info about the damager in the event, since the API states the damager is non-null: 
image

This PR changes canDamage to call the FakeEntityDamageEvent instead if the attacker is null.

Fixes https://github.com/mcMMO-Dev/mcMMO/issues/4767


nossr50:
Thank you

PR #4870, opened 2023-02-25
Fix missing return in notify command

Warriorrrr:
Fixes an exception when a player with no loaded data runs /notify.


nossr50:
Thank you.

PR #4885, opened 2023-04-03
Fix /party teleport tab complete NPE

Warriorrrr:
Fixes an NPE that occurs when a player without a party tries to tab complete after typing /party teleport.


nossr50 (approved):
Thank you

PR #4889, opened 2023-04-22
Fix child skills counting towards power level in /inspect for offline players

Warriorrrr:
Fixes the power level in /inspect for offline players showing higher than their actual power level due to salvage and smelting being counted towards it. It will now iterate over SkillTools#NON_CHILD_SKILLS to calculate the power level similar to what McMMOPlayer#getPowerLevel does.

Originally reported in [EarthMC/Issue-Tracker/issues/1670](https://github.com/EarthMC/Issue-Tracker/issues/1670) which also contains some screenshots of the issue.


nossr50:
Thank you again Warriorrrrr

PR #4895, opened 2023-05-21
Fix blast mining bonus drops option not working

Warriorrrr:
Credit to Mads on the discord for finding this bug, tested it myself aswell to confirm that it's fixed.


nossr50:
Thanks to both of you

PR #4906, opened 2023-06-20
Cleaned up PAPI Expansion

cj89898:
Cleaned up PAPI Expansion, narrowing it down to only 1 class


src/main/java/com/gmail/nossr50/placeholders/PapiExpansion.java
```diff
@@ -49,105 +44,41 @@ public String getRequiredPlugin() {
...
+        return String.valueOf(ExperienceConfig.getInstance().getExperienceGainsGlobalMultiplier());
+      }else if (params.equalsIgnoreCase("power_level_cap")) {
+        return mcMMO.p.getGeneralConfig().getPowerLevelCap()+"";
+      }
+      
+      final McMMOPlayer user = UserManager.getPlayer(player);
+      if (user == null) return null;
+      
+      if (params.startsWith("level_")) {
+        PrimarySkillType skill = PrimarySkillType.valueOf(params.substring(6).toUpperCase());
```
nossr50, line 61: you could use make use of constants to avoid these hard coded magic numbers
public static final String SKILL_LEVEL = "level_";

// later on
PrimarySkillType skill = PrimarySkillType.valueOf(params.substring(SKILL_LEVEL.length()).toUpperCase());

src/main/java/com/gmail/nossr50/placeholders/PapiExpansion.java
```diff
@@ -49,105 +44,41 @@ public String getRequiredPlugin() {
...
+        return skill == null ? null : user.getSkillLevel(skill)+"";
+      }else if (params.startsWith("xp_needed_")) {
```
nossr50, line 63: would prefer space after }

src/main/java/com/gmail/nossr50/placeholders/PapiExpansion.java
```diff
@@ -49,105 +44,41 @@ public String getRequiredPlugin() {
...
+        PrimarySkillType skill = PrimarySkillType.valueOf(params.substring(10).toUpperCase());
+        return skill == null ? null : user.getXpToLevel(skill)+"";
+      }else if (params.startsWith("xp_remaining_")) {
+        PrimarySkillType skill = PrimarySkillType.valueOf(params.substring(13).toUpperCase());
+        return skill == null ? null : (user.getXpToLevel(skill) - user.getSkillXpLevel(skill))+"";
+      }else if (params.startsWith("xp_")) {
+        PrimarySkillType skill = PrimarySkillType.valueOf(params.substring(3).toUpperCase());
+        return skill == null ? null : user.getSkillXpLevel(skill)+"";
```
nossr50, line 71: Would prefer String.valueOf() but this is perhaps a small nitpick

nossr50 (changes requested):
Would you be up for changing it a bit further before I merge it?

src/main/java/com/gmail/nossr50/placeholders/PapiExpansion.java
cj89898: Would you like this done for those only needing substring or all?
Eg. `"in_party"`

nossr50: for all would make sense to me

cj89898:
Should be all tidied up now

Issue #4911, opened 2023-07-08
Packaging mcmmo on windows gives POSIX error on testing stage


Greymagic27:
When using 'mvn clean package install' on windows I get this error when it reaches the testing stage:

https://paste.gg/p/anonymous/9c11a247740d401691841691f81361b9

I then ran the same command with -X to get a log with full debug logging:

https://paste.gg/p/anonymous/4d42fdba5ed0488a99a6e0f3f9db759a

I also ran it with -e to get a full stack trace of the errors:
https://paste.gg/p/anonymous/86beb00a032349d184d03ad15f7e4c90

After some googling it looks like windows does not support 'POSIX' which is causing this error.

Chew:
Seems to be an issue with the surefire plugin? The error is because Windows isn't a POSIX system. It is several versions behind tho (`3.0.0-M7` versus `3.1.2`)

Greymagic27:
Updated to 3.1.2 on my fork however I still have the same issue.

However, it has given me a specific file now: https://paste.gg/p/anonymous/0b4e844ba8e14e109dc0aeb72eafe037

Demon-tk:
Have the same issue https://gist.github.com/Demon-tk/9b91c65c243189dde4bb1c5f8461a33b

Chew:
Correct me if I'm wrong but doesn't Windows have some way to run Linux natively now? But again, this is an issue with whatever maven plugin that is, or if it's intended, I'm sure it's documented somewhere and fixable on your end. The best we can do is figure out a different plugin for what we need.

Greymagic27:
Windows can run linux through WSL, but I don't think it should be a requirement to compile the plugin, it never was before.

From both my logs and the other users, the error seems to be being produced from https://github.com/mcMMO-Dev/mcMMO/blob/master/src/test/java/com/gmail/nossr50/database/FlatFileDatabaseManagerTest.java. Specifically, the error points out line 78, which is 'tempDir = Files.createTempDir();'. This is causing the error as it's not windows compatible code to my best understanding.

Reading https://itecnote.com/tecnote/java-lang-unsupportedoperationexception-posixpermissions-not-supported-as-initial-attribute-on-windows/, it seems like a similar issue.

Regarding fixable on our end: This an issue with the plugin not having file support for windows, not a problem on our respective PCs.

catronomix:
Tried to build today and get the same error

milizm:
[...]

PR #4919, opened 2023-07-30
Fix packaging error on windows #4911

Greymagic27:
This fixes https://github.com/mcMMO-Dev/mcMMO/issues/4911

Tested on Windows 11 locally using 'mvn clean package install'. Console output not included due to containing personal information (PC username), however can comfirm this solves the POSIX issue.


PatalJunior:
Can confirm it works as well

nossr50:
Thank you

PR #4925, opened 2023-08-17
mcMMO-Folia-PR

Yomamaeatstoes:
Here is the PR for mcMMO Folia. We are still yet to replace all the schedulers. Although here is the progress we have so far. Here is the library we have been using for this. https://github.com/TechnicallyCoded/FoliaLib


nossr50:
Impressive guys, great job.

PR #4934, opened 2023-09-01
add access to player in McMMoPlayerNotificationEvent

JL-III:
Can we add access to the player in this notification event?


nossr50:
I went ahead and added this myself in 2.1.225

PR #4935, opened 2023-09-03
update FoliaLib

HSGamer:
- fix an issue with 0 ticks setting
- disable scheduler warning


src/main/java/com/gmail/nossr50/mcMMO.java
```diff
@@ -173,6 +174,7 @@ public void onEnable() {
 
             //Folia lib plugin instance
             foliaLib = new FoliaLib(this);
+            InvalidTickDelayNotifier.disableNotifications = true;
```
nossr50, line 177: I don't understand this change?

nossr50 (changes requested):
Turning the warnings off doesn't seem like a good idea to me, could you explain or change it?

nossr50:
I read up on it, apparently the lib will convert 0 tick tasks automatically, so this is fine.

PR #4942, opened 2023-10-03
General Fixes and Cleanup for Folia Support

### LINKED ISSUE #4930: [Folia] Could not pass event McMMOPlayerLevelUpEvent
mcMMO v2.1.223
Folia (fork)
https://mclo.gs/S3Lg3yR

TechnicallyCoded:
**Description**
Fixes, Cleanup, and Comments for actions that need further investigation

**Issues**
Fixes #4930

**Notes**
Ctrl + F -> "// TODO TECH CHECK" to find sections that still need work before I consider this done 😄

### DIFF
--- a/src/main/java/com/gmail/nossr50/commands/party/teleport/PtpAcceptCommand.java
+++ b/src/main/java/com/gmail/nossr50/commands/party/teleport/PtpAcceptCommand.java
@@ -49,7 +49,7 @@ public class PtpAcceptCommand implements CommandExecutor {
         }
 
         if (mcMMO.p.getGeneralConfig().getPTPCommandWorldPermissions()) {
-            // TODO TECH CHECK - running operations on teleporting player and target player without the proper thread
+            // TODO FOLIA - target is not on the same thread as sender/player - unsafe
             World targetWorld = target.getWorld();
             World playerWorld = player.getWorld();
 
@@ -65,7 +65,7 @@ public class PtpAcceptCommand implements CommandExecutor {
             }
         }
 
-        // TODO TECH CHECK - running operations on teleporting player and target player without the proper thread
+        // TODO FOLIA - target is not on the same thread as sender/player - unsafe
         PtpCommand.handleTeleportWarmup(target, player);
         return true;
     }
--- a/src/main/java/com/gmail/nossr50/commands/party/teleport/PtpAcceptCommand.java
+++ b/src/main/java/com/gmail/nossr50/commands/party/teleport/PtpAcceptCommand.java
@@ -49,6 +49,7 @@ public class PtpAcceptCommand implements CommandExecutor {
         }
 
         if (mcMMO.p.getGeneralConfig().getPTPCommandWorldPermissions()) {
+            // TODO TECH CHECK - running operations on teleporting player and target player without the proper thread
             World targetWorld = target.getWorld();
             World playerWorld = player.getWorld();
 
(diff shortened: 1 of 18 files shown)


TechnicallyCoded:
More major design decisions that still need work under Folia are now marked with "// TODO FOLIA"

src/main/java/com/gmail/nossr50/commands/party/teleport/PtpAcceptCommand.java
```diff
@@ -49,6 +49,7 @@ public boolean onCommand(@NotNull CommandSender sender, @NotNull Command command
         }
 
         if (mcMMO.p.getGeneralConfig().getPTPCommandWorldPermissions()) {
+            // TODO FOLIA - target is not on the same thread as sender/player - unsafe
```
nossr50, line 52: I would like these "unsafe" operations to be addressed, or at least an explanation for why they cannot be addressed.

src/main/java/com/gmail/nossr50/listeners/WorldListener.java
```diff
@@ -29,9 +31,16 @@ public void onStructureGrow(StructureGrowEvent event) {
...
+        // Under Folia, any two loaded and adjacent chunks will be in the same ticking region. To avoid scheduling many
+        // tasks, we will use this assumption.
+        // Therefore, we will schedule the task such that all blocks in the event are handled from the ticking region of
+        // the first block in the event.
+        // Without folia, this will run on the main tick.
+        List<BlockState> blocks = event.getBlocks();
+        BlockState referenceBlock = blocks.get(0);
+        plugin.getFoliaLib().getImpl().runAtLocationLater(referenceBlock.getLocation(), () -> {
+            for (BlockState blockState : blocks) {
+                // The HashChunkManager is thread-safe, we can safely call it from a non-single-threaded environment
```
nossr50, line 43: This is not true

nossr50 (changes requested):
Some things need to be changed and addressed.

src/main/java/com/gmail/nossr50/commands/party/teleport/PtpAcceptCommand.java
TechnicallyCoded: This was already briefly discussed in Discord DMs. I don't really remember everything I looked at as it's been a while ^^. Most likely due to this method, or an upstream method requiring/expecting immediate returns of data that would necessitate both ticking regions to be synchronized. This is obviously not how Folia works but would require a re-write of some major parts under mcMMO to solve.

To clarify, these issues were already present. I just commented them as potentially troublesome for future reference and maintenance. The PR is not focused on fixing absolutely everything under folia but rather improving some blatantly wrong parts. I also happened to add comments to stuff that could be an issue going forward.

src/main/java/com/gmail/nossr50/listeners/WorldListener.java
TechnicallyCoded: I will review this again, as stated above, I don't remember everything I had checked back then. It's been a while.

TechnicallyCoded: I have no idea why I thought this was thread-safe. I do in-fact agree. However, simply changing that class to using concurrent maps could introduce side-effects that would go further than just folia. It's probably better to stick with a more risky folia impl than to risk messing up the stable releases under normal bukkit.

I can remove the comment or compeletely revert that change. However, this chunk manager is already being access unsafely by other event listeners under folia and therefore isn't synced to any particular thread under a Folia environment.

nossr50: The proper solution is most likely changing the HashChunkManager to be thread safe, however, that would be a complex task and require a lot of testing. Hearing that it's already being used unsafely makes sense and is unfortunate, this should be addressed, or Folia server owners will run into some problems.

PR #4951, opened 2023-10-26
Fix build failure from missing dependency

PikaMug:
WorldGuardManager.java has `com.sk89q.worledit.bukkit.*` imports which are not included with just worldguard-core, so any new environment will have build failure. Adding worldedit-bukkit to pom.xml resolves this.


nossr50:
Thank you 😊

PR #4976, opened 2024-02-21
Fix exceptions on EntityCombustByEntityEvent, EntityDamageByEntityEvent, PartyAutoKickTask if parties are disabled

IAISI:
As alternative solution it might make sense to run check if parties are disabled in EntitiyListener#checkParties function instead of before running EntitiyListener#checkParties but I'm not sure how verbose we wanna make this, in terms of logic it probably wouldn't make sense to run party related stuff if parties are disabled (potentially make code less readable).


nossr50:
thanks

PR #4981, opened 2024-03-02
Fixed NPE when party creation doesn't use password.  Solves bug repor…

Momshroom:
…ted here: https://discord.com/channels/526933440214597677/526938425161416727/1213131451235827753


nossr50:
Thank you!

PR #5000, opened 2024-04-28
Initiated fix so it at least load on 1.20.5

broken1arrow:
This is only a beginning of tests it out on 1.20.5, PotionData is gone completely and ARROW_INFINITE is called only INFINITE. 

The enchantment could be solved with either version check or try catch. About the PotionData I think the plan is you should use  PotionType class. don't know how new that class is and if should be a wrapper class. If it is easier to split up the extended and upgraded in the config than they have to type full name? could also be optional between full name or the short name.


nossr50:
I appreciate the help, but this is not a proper fix for 1.20.5, I am in the midst of working on it myself, so stay tuned.

PR #5012, opened 2024-05-21
Fix sapling drops to ignore reduction and obey Knock on Wood

Momshroom:
The prior fix ignored KnockOnWood and also checked if the block that was dropping things was a sapling, rather than if the drop was.   This fixes the issue so that saplings beyond the 25% of blocks will drop if KnockOnWood is unlocked.

(I don't know how to remove the two extra commits from the PR, sorry, but they do cancel each other out)


src/main/java/com/gmail/nossr50/util/Misc.java
```diff
@@ -128,6 +128,20 @@ public static void spawnItemsFromCollection(@Nullable Player player, @NotNull Lo
...
+    /**
+     * Drops the item from the item stack only if it is a sapling (or equivalent)
+     * Needed for TreeFeller
+     */
+    public static void spawnItemIfSapling(@NotNull Player player, @NotNull Location location, @NotNull Collection < ItemStack > drops, @NotNull ItemSpawnReason itemSpawnReason) {
+        String dropString;
+        for (ItemStack drop : drops) {
+            dropString = drop.getType().getKey().getKey();
+            if (dropString.contains("sapling") || dropString.contains("propagule")) {
+                Misc.spawnItem(player, location, drop, itemSpawnReason);
```
nossr50, line 140: you're already in Misc so you can drop `Misc.`

src/main/java/com/gmail/nossr50/util/Misc.java
```diff
@@ -128,6 +128,20 @@ public static void spawnItemsFromCollection(@Nullable Player player, @NotNull Lo
         }
     }
 
+    /**
+     * Drops the item from the item stack only if it is a sapling (or equivalent)
+     * Needed for TreeFeller
+     */
+    public static void spawnItemIfSapling(@NotNull Player player, @NotNull Location location, @NotNull Collection < ItemStack > drops, @NotNull ItemSpawnReason itemSpawnReason) {
```
nossr50, line 135: A few notes...
1. What does passing `player` in contribute to this method? (actually, what is it doing in `spawnItem`?)
2. I would rename `location` to `spawnLocation` for clarity
3. Remove the extra whitespace in `< ItemStack >`
4. This method could be made more generic, for example something like this
```conditionallySpawn(@NotNull Predicate<String> predicate, @NotNull Location spawnLocation, @NotNull Collection<ItemStack> items, @NotNull ItemSpawnReason itemSpawnReason)```

**Review by nossr50 — CHANGES_REQUESTED** [review]

File: src/main/java/com/gmail/nossr50/util/Misc.java [hunk 3]
```diff
@@ -128,6 +128,20 @@ public static void spawnItemsFromCollection(@Nullable Player player, @NotNull Lo
...
+        String dropString;
```
nossr50, line 136: You should move this variable declaration to the same line it is initialized, you can also make it ```final```

src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java
```diff
@@ -322,13 +322,17 @@ private void dropTreeFellerLootFromBlocks(@NotNull Set<BlockState> treeFellerBlo
...
                 // 75% of the time do not drop leaf blocks
-                if (blockState.getType().getKey().getKey().toLowerCase().contains("sapling")
-                        || ThreadLocalRandom.current().nextInt(100) > 75) {
+                if (ThreadLocalRandom.current().nextInt(100) > 75) {
                     Misc.spawnItemsFromCollection(getPlayer(),
                             Misc.getBlockCenter(blockState),
                             block.getDrops(itemStack),
                             ItemSpawnReason.TREE_FELLER_DISPLACED_BLOCK);
                 }
+                // drop saplings as occur in rest if Knock on Wood unlocked.
```
nossr50, line 331: grammar/spelling is odd here

src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java
```diff
@@ -322,13 +322,17 @@ private void dropTreeFellerLootFromBlocks(@NotNull Set<BlockState> treeFellerBlo
...
+                else if (RankUtils.hasUnlockedSubskill(player, SubSkillType.WOODCUTTING_KNOCK_ON_WOOD)) {
+                    Misc.spawnItemIfSapling(getPlayer(), Misc.getBlockCenter(blockState),
```
nossr50, line 333: if ```player``` is already in scope, you don't need ```getPlayer()``` and ```player``` combined, they are the same reference

src/main/java/com/gmail/nossr50/util/Misc.java
Momshroom: lol. will do.  I wasn't in Misc at the point I wrote that bit xD

src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java
Momshroom: fixed in upcoming addition to PR.

src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java
Momshroom: Fixing there and in the existing code above it.

src/main/java/com/gmail/nossr50/util/Misc.java
Momshroom: The code snippet shown doesn't have a variable declaration shown.   
If you mean dropString, it is reset inside the for loop, so declaring it where it is initialized would require the variable to be redone every loop, which is why I declared outside the loop and why it isn't final, so I'm guessing you meant to link different code, but I'm not sure what else you might mean to fix it?

src/main/java/com/gmail/nossr50/util/Misc.java
Momshroom: 1) it's in there because I need to pass it to spawnItem.   It appears to be in spawnItem because you use it in the McMMOItemSpawnEvent constructor on line 172

2) Renamed.  It was location to match your spawnItem method.

3) Done.

4). Will figure out how to do this and resubmit.  It seemed overly specific as written.  Ty.

src/main/java/com/gmail/nossr50/util/Misc.java
Momshroom: Ah, here I see you do mean drop string.  I didn't move it because it needs to be reset as the drops are looped.

src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java
Momshroom: fixed

src/main/java/com/gmail/nossr50/skills/woodcutting/WoodcuttingManager.java
Momshroom: fixed

src/main/java/com/gmail/nossr50/util/Misc.java
Momshroom: 4) now done

src/main/java/com/gmail/nossr50/util/Misc.java
Momshroom: variable removed entirely now that using predicate

src/main/java/com/gmail/nossr50/util/Misc.java
Momshroom: done

nossr50 (approved):

PR #5024, opened 2024-06-18
Fix exceptions when shooting entities through worlds on folia

Warriorrrr:
The AwardCombatXpTask checks the target's health to see how much damage was done, but running this on the player's scheduler can cause exceptions if the target is not in the same region, like if the target is on the other side of a nether portal.

### DIFF
--- a/src/main/java/com/gmail/nossr50/skills/taming/TamingManager.java
+++ b/src/main/java/com/gmail/nossr50/skills/taming/TamingManager.java
@@ -287,6 +287,9 @@ public class TamingManager extends SkillManager {
         double range = 5;
         Player player = getPlayer();
 
+        if (!target.getWorld().equals(player.getWorld()))
+            return;
+
         for (Entity entity : player.getNearbyEntities(range, range, range)) {
             if (entity.getType() != EntityType.WOLF) {
                 continue;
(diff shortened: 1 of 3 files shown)


src/main/java/com/gmail/nossr50/skills/taming/TamingManager.java
```diff
@@ -287,6 +287,9 @@ public void attackTarget(LivingEntity target) {
         double range = 5;
         Player player = getPlayer();
 
+        if (!target.getWorld().equals(player.getWorld()))
```
nossr50, line 290: When would a player's world not equal their targets?

nossr50 (commented):
Left a question, sorry for the late review.

src/main/java/com/gmail/nossr50/skills/taming/TamingManager.java
Warriorrrr: It's possible to shoot an arrow through a nether portal and hit an entity on the other side of it

nossr50: Thank you for that information!

nossr50 (approved):
Thank you!

PR #5038, opened 2024-07-03
Fix mace skills permission

### LINKED ISSUE #5037: Permissions Maces
You won't get any xp from maces while user have the perm mcmmo.defaults

the permission mcmmo.skills.maces is required to gain xp.

Please make maces part of the default perms group.

SrBedrock:
Close #5037


nossr50:
Good catch, thank you!

PR #5071, opened 2024-08-16
Now firing events before changing durability

Mowstyl:
By calling the PlayerItem(Damage|Mend)Event before setting the durability, plugins that handle the durability of items themselves (like any custom items plugin) won't break.


src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
```diff
@@ -236,16 +241,50 @@ public static void handleDurabilityChange(ItemStack itemStack, int durabilityMod
...
+        // Call PlayerItemDamageEvent first to make sure it's not cancelled
+        if (damage > 0) {  // Damaged item
+            PlayerItemDamageEvent event = new PlayerItemDamageEvent(player, itemStack, finalDurability);
+            Bukkit.getPluginManager().callEvent(event);
+            if (!event.isCancelled() && event.getDamage() != 0) {
+                itemStack.setDurability((short) (itemStack.getDurability() + event.getDamage()));
+            }
+        } else if (damage < 0) {  // Repaired item
+            damage = -damage;
+            ExperienceOrb auxOrb = (ExperienceOrb) player.getWorld().spawnEntity(player.getLocation(), EntityType.EXPERIENCE_ORB);
```
nossr50, line 260: What's this experience orb stuff about? I don't get it.
Doesn't seem related to your PR.

nossr50 (changes requested):
I left some feedback, reply to it when you get a chance.

src/main/java/com/gmail/nossr50/util/skills/SkillUtils.java
Mowstyl: If the damage is negative, that means the item has been repaired. The PlayerItemMendEvent has an ExperienceOrb as an argument, so I needed to spawn one with the quantity of xp needed to repair the item for the specified amount if you had the mending enchantment. After the event has been called, the xp orb has to be despawned so no one can pick it up.
After your review I noticed I forgot to call the event, I just instantiate it and check if it has been canceled.

Mowstyl: I thought it would be better to call the mend event instead of the damage event when repairing an item, yet I think everything would also work with a damage event. I just did that for the sake of semantics.

nossr50: I see, thank you for the explanation, I will think on that some more.

PR #5075, opened 2024-08-25
Cache ItemMeta instead of creating a new ItemMeta for each potion check

MrPowerGamerBR:
The `isSimilarPotion` is very expensive because the code attempts to get the `otherPotion` (in this case, the current item) `ItemMeta` on every comparison check, and initializing the `ItemMeta` is *very* expensive.

To avoid this, we only create the `ItemMeta` once and then reuse it in the `isSimilarPotion` check.

While on Spigot the ItemStack code does seem to cache the ItemMeta, on Paper it doesn't due to the ItemStack being delegated to the server ItemStack, the ItemMeta seems to be always be created on every `getItemMeta()` call.

image

### DIFF
--- a/src/main/java/com/gmail/nossr50/config/skills/alchemy/PotionConfig.java
+++ b/src/main/java/com/gmail/nossr50/config/skills/alchemy/PotionConfig.java
@@ -335,6 +335,10 @@ public class PotionConfig extends LegacyConfigLoader {
      * @return AlchemyPotion that corresponds to the given ItemStack.
      */
     public AlchemyPotion getPotion(ItemStack item) {
+        // Fast return if the item does not have any item meta to avoid initializing an unnecessary ItemMeta instance
+        if (!item.hasItemMeta())
+            return null;
+
         ItemMeta itemMeta = item.getItemMeta();
         final List<AlchemyPotion> potionList = alchemyPotions.values()
                 .stream()
--- a/src/main/java/com/gmail/nossr50/config/skills/alchemy/PotionConfig.java
+++ b/src/main/java/com/gmail/nossr50/config/skills/alchemy/PotionConfig.java
@@ -9,6 +9,7 @@ import org.bukkit.Color;
 import org.bukkit.Material;
 import org.bukkit.configuration.ConfigurationSection;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.ItemMeta;
 import org.bukkit.inventory.meta.PotionMeta;
 import org.bukkit.potion.PotionEffect;
 import org.bukkit.potion.PotionEffectType;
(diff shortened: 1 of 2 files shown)


src/main/java/com/gmail/nossr50/datatypes/skills/alchemy/AlchemyPotion.java
```diff
@@ -18,13 +19,15 @@
...
     private final @NotNull ItemStack potionItemStack;
+    private final @NotNull ItemMeta potionItemMeta;
     private final @NotNull Map<ItemStack, String> alchemyPotionChildren;
 
     public AlchemyPotion(@NotNull String potionConfigName, @NotNull ItemStack potionItemStack,
                          @NotNull Map<ItemStack, String> alchemyPotionChildren) {
         this.potionConfigName = requireNonNull(potionConfigName, "potionConfigName cannot be null");
         this.potionItemStack = requireNonNull(potionItemStack, "potionItemStack cannot be null");
         this.alchemyPotionChildren = requireNonNull(alchemyPotionChildren, "alchemyPotionChildren cannot be null");
+        this.potionItemMeta = potionItemStack.getItemMeta(); // The potion item meta should never be null because it is a potion
```
nossr50, line 30: add `requireNonNull`

src/main/java/com/gmail/nossr50/config/skills/alchemy/PotionConfig.java
```diff
@@ -334,9 +335,14 @@ public AlchemyPotion getPotion(String name) {
      * @return AlchemyPotion that corresponds to the given ItemStack.
      */
     public AlchemyPotion getPotion(ItemStack item) {
+        // Fast return if the item does not have any item meta to avoid initializing an unnecessary ItemMeta instance
```
nossr50, line 338: Smart, I like this change.

src/main/java/com/gmail/nossr50/datatypes/skills/alchemy/AlchemyPotion.java
```diff
@@ -49,21 +52,26 @@ public AlchemyPotion(@NotNull String potionConfigName, @NotNull ItemStack potion
...
+    public boolean isSimilarPotion(@NotNull ItemStack otherPotion, @Nullable ItemMeta otherMeta) {
         requireNonNull(otherPotion, "otherPotion cannot be null");
+
         if (otherPotion.getType() != potionItemStack.getType()) {
             return false;
         }
 
         // no potion meta, no match
-        if (!otherPotion.hasItemMeta()) {
+        if (otherMeta == null) {
```
nossr50, line 66: Don't understand why this change is necessary, isn't it equivalent to the prior version?

nossr50 (changes requested):
Looks good, small changes requested.

src/main/java/com/gmail/nossr50/datatypes/skills/alchemy/AlchemyPotion.java
MrPowerGamerBR: It is equivalent, but iirc I added that check because IDEA complains that you are using a maybe null value in `hasDifferingCustomEffects`... or maybe it was my Kotlin brain saying that I should add a null check.

MrPowerGamerBR: image

nossr50:
Thank you, I appreciate it!

PR #5088, opened 2024-09-20
Fix acrobatics damage calculations

Ineusia:
Currently, acrobatics does it's damage calculates pre-vanilla damage reductions which leads to a non-fatal fall being considered fatal. 

This leads to armored players never procing Roll or Graceful Landing when they infact should, as their fall damage was not fatal.

This change uses the getFinalDamage() method instead of getDamage() method which will use the post-vanilla reduction damage calculation to calculate fatality as well as mcMMO's further reductions.

Using setDamage would re-trigger vanilla damage reductions and would result in a double-reduction, both pre and post mcMMO calculations, so we instead setDamage to 0 from this fall and set a new damage with the DamageType of MAGIC which bypasses resistances.

This is perfectly fine because vanilla reductions are already factored into the fall before mcMMO does any modification, and the server will never see a death from "magic" since Roll will not run any modifications should the fall be fatal.


nossr50:
Thank you

PR #5089, opened 2024-09-20
Fix slime spawner exploit

Ineusia:
Currently, slime/magma cube meta gets cleaned at death which fires before the Transform event. This means the smaller slimes/magma cubes from a spawner slime will not have tags carry over, allowing players to farm combat skills from spawned mobs.

This PR moves cleaning for Slimes/Magma Cubes to after the transform event, which we know means the slime is dead anyways, so it's fine to clean them there.

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/EntityListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/EntityListener.java
@@ -46,6 +46,7 @@ import org.bukkit.potion.PotionEffectType;
 import org.bukkit.projectiles.ProjectileSource;
 
 import java.util.Arrays;
+import java.util.List;
 
 import static com.gmail.nossr50.util.MobMetadataUtils.*;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -652,7 +659,14 @@ else if (livingEntity instanceof Tameable pet) {
      */
     @EventHandler(priority = EventPriority.LOWEST)
     public void onEntityDeathLowest(EntityDeathEvent event) {
-        mcMMO.getTransientMetadataTools().cleanLivingEntityMetadata(event.getEntity());
+        LivingEntity entity = event.getEntity();
+
+        // Clear metadata for Slimes/Magma Cubes after transformation events take place, otherwise small spawned slimes will not have any tags
+        if (Arrays.asList(EntityType.SLIME, EntityType.MAGMA_CUBE).contains(entity.getType())) {
```
nossr50, line 665: Make Arrays.asList(EntityType.SLIME, EntityType.MAGMA_CUBE) a static constant somewhere (could be this class or in somke util) and use that instead

nossr50 (changes requested):

src/main/java/com/gmail/nossr50/listeners/EntityListener.java
```diff
@@ -57,6 +58,7 @@ public class EntityListener implements Listener {
      * check if a {@link Player} has a {@link Trident} enchanted with "Piercing".
      */
     private final NamespacedKey piercingEnchantment = NamespacedKey.minecraft("piercing");
+    private final List<EntityType> transformableEntities = Arrays.asList(EntityType.SLIME, EntityType.MAGMA_CUBE);
```
nossr50, line 61: This should be static as well, I'll make the change after I merge.

PR #5094, opened 2024-09-27
Bugfix/acrobatics roll chance

Ineusia:
Moves the roll damage reduction into the block that executes if roll was successful.


nossr50:
Thank you

PR #5095, opened 2024-09-27
Redo XP perks

Jsinco:
This PR removes the old 'else if' chain used for xp perks and replaces it with an enum class for faster usage and other benefits. This PR also brings another important change by ensuring  the **highest** XP multiplier amount is always used for the player.

Before this change, mcMMO would always use the custom multiplier if the player had permission for it even if it was lower than other multiplier amounts. This would cause conflicts for those who want to use the custom xp amount for a lower xp boost, and then give their players higher boost amounts later on.

### DIFF
--- a/src/main/java/com/gmail/nossr50/util/skills/XPBoostAmount.java
+++ b/src/main/java/com/gmail/nossr50/util/skills/XPBoostAmount.java
@@ -1,0 +1,94 @@
+package com.gmail.nossr50.util.skills;
+
+import com.gmail.nossr50.config.experience.ExperienceConfig;
+import com.gmail.nossr50.datatypes.skills.PrimarySkillType;
+import org.bukkit.permissions.Permissible;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+public final class XPBoostAmount {
+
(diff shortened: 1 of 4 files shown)


nossr50:
I appreciate this PR but I'll have to think about if I merge this or not, you see I have a complete rewrite for perks in another branch that adds new functionality, I'd just have to port the code over... 😮‍💨

Jsinco:
It would probably be better to either:

A. Merge and fix this issue
B. Fix this yourself

I don't think my solution is bad, or even hard to reverse if needed. But leaving this unpatched shouldn't really be considered?

nossr50:
Good points, I'll think on it, and review the PR in the near future.

src/main/java/com/gmail/nossr50/util/skills/XPBoostAmount.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    public String name() {
+        return FIELD_NAME;
+    }
+
+    public static final Map<String, XPBoostAmount> VALUES = new HashMap<>();
+    public static final List<XPBoostAmount> VALUES_SORTED_BY_MULTIPLIER = new ArrayList<>(); // Faster to just sort once since mcMMO doesn't have any reloading anyway
+
+    static {
```
nossr50, line 68: Use a defined list instead of reflection

PR #5098, opened 2024-09-28
Add Permission forget

Euphillya:
/mmodebug : mcmmo.commands.mmodebug
/mmocompat : mcmmo.commands.mmocompat
/mmoxpbar : mcmmo.commands.mmoxpbar


nossr50:
All three of these purposely do not have permissions as they should always be available.

PR #5111, opened 2024-11-13
add McMMOPlayerMasterAnglerEvent

bobcat4848:
Added a new event: McMMOPlayerMasterAnglerEvent so that developers who interface can cancel this event if they desire.

Super simple implementation, tested and working as expected.


nossr50:
you weren't kidding when you said super simple, this time I'll extend it after pulling it in 👍

PR #5141, opened 2025-01-18
Remove globally affecting call to Locale#setDefault

BlvckBytes:
Greetings!

The call to `Locale#setDefault` within `LocaleLoader#initialize` threw me for quite the massive loop today, as it affected multiple other plugins which assumed `Locale#getDefault` to be the JVM's, and thus the system's locale, which mcMMO hard-coded to `en_US`.

While I have to admit that those other developers don't quite adhere to best practises by depending on this global state either, I still wholeheartedly believe that mcMMO should refrain from altering JVM-global-state nonetheless.

Also, since this initialization may occur after executing some of the `static`-blocks of other classes containing various formatters, many test-cases failed to execute due to formatting-discrepancies on my local machine, having had a non-english locale configured.

My proposal represents merely a quick patch, as to cut out altering global state, but the ideal solution would obviously be to use the locale as configured on the plugin; that would demand a few architectural changes, which I do not quite have the resources for, at this point in time.

I hope that this tiny but still rather significant issue can be settled soon.

Best Regards,
BlvckBytes

### DIFF
--- a/src/main/java/com/gmail/nossr50/commands/skills/SkillCommand.java
+++ b/src/main/java/com/gmail/nossr50/commands/skills/SkillCommand.java
@@ -25,6 +25,7 @@ import org.bukkit.entity.Player;
 import org.jetbrains.annotations.NotNull;
 
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


nossr50:
Thank you for the write up and thorough explanation, and thank you for the PR as well.

PR #5142, opened 2025-01-19
Add event for players taming an entity

Warriorrrr:
Adds an event that allows plugins to modify xp gain for taming specific entities, or to cancel the xp gain outright.

### DIFF
--- a/src/main/java/com/gmail/nossr50/events/skills/taming/McMMOPlayerTameEntityEvent.java
+++ b/src/main/java/com/gmail/nossr50/events/skills/taming/McMMOPlayerTameEntityEvent.java
@@ -1,0 +1,67 @@
+package com.gmail.nossr50.events.skills.taming;
+
+import com.gmail.nossr50.datatypes.experience.XPGainReason;
+import com.gmail.nossr50.datatypes.skills.PrimarySkillType;
+import com.gmail.nossr50.events.experience.McMMOPlayerExperienceEvent;
+import com.google.common.base.Preconditions;
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Called when a player has tamed an entity, and we are about to award them experience for it.
+ */
+public class McMMOPlayerTameEntityEvent extends McMMOPlayerExperienceEvent {
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private float xpGained;
+    private final Entity tamedEntity;
+
+    @ApiStatus.Internal
+    public McMMOPlayerTameEntityEvent(@NotNull Player player, float xp, @NotNull Entity tamedEntity) {
+        super(player, PrimarySkillType.TAMING, XPGainReason.PVE);
+        this.xpGained = xp;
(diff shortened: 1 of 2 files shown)


src/main/java/com/gmail/nossr50/events/skills/taming/McMMOPlayerTameEntityEvent.java
```diff
@@ -0,0 +1,67 @@
...
+    public float getXpGained() {
+        return this.xpGained;
+    }
+
+    /**
+     * @param xpGained The raw experience that the player will receive.
+     * @throws IllegalArgumentException if xpGained is NaN or infinite.
+     */
+    public void setXpGained(float xpGained) {
+        Preconditions.checkArgument(Float.isFinite(xpGained), "new gained xp must be a number");
```
nossr50, line 41: You shouldn't need this check since you are passing in a primitive type, instead it would be a good idea to make sure the float is a positive number

src/main/java/com/gmail/nossr50/events/skills/taming/McMMOPlayerTameEntityEvent.java
```diff
@@ -0,0 +1,67 @@
...
+ * Called when a player has tamed an entity, and we are about to award them experience for it.
+ */
+public class McMMOPlayerTameEntityEvent extends McMMOPlayerExperienceEvent {
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private float xpGained;
+    private final Entity tamedEntity;
+
+    @ApiStatus.Internal
+    public McMMOPlayerTameEntityEvent(@NotNull Player player, float xp, @NotNull Entity tamedEntity) {
```
nossr50, line 23: I would expect the `McMMOPlayer` to be dependency injected into this event, with a getter (no setter).

src/main/java/com/gmail/nossr50/events/skills/taming/McMMOPlayerTameEntityEvent.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Called when a player has tamed an entity, and we are about to award them experience for it.
+ */
+public class McMMOPlayerTameEntityEvent extends McMMOPlayerExperienceEvent {
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    private float xpGained;
+    private final Entity tamedEntity;
+
+    @ApiStatus.Internal
```
nossr50, line 22: Not only do we not make use of `ApiStatus.Internal` anywhere else, this being a public ctor is VERY much a public API... so my recommendation is you just remove this annotation

nossr50 (changes requested):
Some small changes needed

src/main/java/com/gmail/nossr50/events/skills/taming/McMMOPlayerTameEntityEvent.java
Warriorrrr: I've added a check for positive numbers, but the isFinite one is still useful because a primitive float can be NaN/infinite

nossr50: @Warriorrrr Thank you for hat info, it seems you are correct on this, my mistake!

nossr50 (approved):

PR #5173, opened 2025-04-19
Fix spectral arrows

broccolai:
Currently spectral arrows don't give crossbow or archery xp

SpectralArrow does not extend Arrow only AbstractArrow, AbstractArrow can be an Arrow, SpectralArrow or a Trident. Trident is already checked in a previous statement so this should be fine


nossr50:
Thank you

Issue #5175, opened 2025-04-25
Blastmining Mob Spawner drop


Seraphoenixde:
Blastmining allows players to get mobspawners.
Place tnt 4 blocks from the spawner, get some distance (2 chunks or more) and prime the tnt with your pickaxe.
The mobspawner drops as an item. This shouldn't happen.

mcmmo version: 2.2.035
paper: 1.21.4-222

nossr50:
I'm unable to replicate this in testing.
If you could provide more details, that might help.
I tried with and without silk touch pickaxe.

Using
`[14:07:56 INFO]: Checking version, please wait...
[14:07:56 INFO]: This server is running Paper version 1.21.4-226-ver/1.21.4@a838a88 (2025-04-17T19:20:57Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
You are running the latest version
Previous version: 1.21.4-212-bb1beda (MC: 1.21.4)`

Seraphoenixde:
[This video should help](https://www.youtube.com/watch?v=2RZePLuPEf0)
This doesn't happen 100% of the time, I'm not 100% sure what the correct setup is.

Tnt needs distance to the spawner, like 3 blocks. Ignition with the pickaxe from more than 1 chunk seems to be pretty reliable to replicate this.

nossr50:
Thanks, I have merged that PR.

PR #5178, opened 2025-05-05
Added spawners to illegal drops for blast mining

TomBock:
Related to: https://github.com/mcMMO-Dev/mcMMO/issues/5175

Spawners had a 10% chance of dropping when exploded with TNT (like most non-ores). Now they no longer drop


nossr50:
Thank you sir, good catch.

PR #5194, opened 2025-07-09
Fixed a bug that caused empty buckets to burn. Fix #4940:

### LINKED ISSUE #4940: Smelting bug
When I'm smelting minerals (no food), the furnace uses bucket as fuel. This bug dont appear when I unloaded mcmmo. So I think this is a bug of mcmmo. Could u solve it or give me a solution?

I am using the version 2.1.202. But I still dont't see any related update in 2.1.204 and 2.1.203.

RunqRun:
https://github.com/mcMMO-Dev/mcMMO/issues/4940
Empty buckets can be placed in the FURNACE with vanilla specifications.
An empty bucket has a burnTime of 0, but if you put 0 in the fuelEfficiency method, 1 will be returned.
If burnTime is then set to 1 by the setBurnTime method, the software will consume the empty bucket with a burnTime of 1.


nossr50:
Thank you @RunqRun

PR #5201, opened 2025-07-21
(Improvement) Implement playing sound by string ID

JeBobs:
I've replaced enum-based sound playing events with string-based equivalents, which should open the door for server customization and other enhancements in the future

- Added SoundLookup class with different registry lookup methods depending on server version.
- Added the ability to configure what sounds are played depending on event, with a fallback built into SoundType.
- Removed getCrippleSound as SoundLookup can now fall back to the original default sound if the mace sound doesn't exist on the server's Minecraft version.
- Added a EnableCustomSounds config variable that will skip SoundLookup ID checking and just pass the sound string directly to the client, mainly due to the fact that it isn't possible to verify if resource pack values exist.
 - Cleaned up a few switch statements to match how the original getSound had it formatted.

I'd love to see/do a further expansion of sound configuration for each ability now that we can just fall back to generic, but that may be for another PR.


nossr50:
@JeBobs have you test on Spigot and Paper?

nossr50:
@JeBobs Don't mind the copilot comments too much, I saw the "request review from copilot" and was wondering what the heck it would even do, so I pressed it to "try it and see".

JeBobs:
i tested spigot 1.20.4 and paper 1.21.4 as those are what I've had on hand, but I'll do more extensive testing later today

nossr50:
> i tested spigot 1.20.4 and paper 1.21.4 as those are what I've had on hand, but I'll do more extensive testing later today

Mostly, I'd want you to test Spigot and Paper for the oldest supported version as well to make sure no exceptions are being thrown / etc.

As of right now, that would be Minecraft version `1.18.2`

nossr50:
I'll make some tweaks to your branch and merge afterwards, might not get to it this weekend.

nossr50:
@JeBobs I'm making some optimizations now, reflection calls are expensive and should be cached. I'm also changing how custom sounds are defined, you'll see soon.

PR #5212, opened 2025-08-24
Modify itemstack directly instead of spawning a new one

GriffinCodes:
Ran into an issue on my server where I couldn't modify items spawned by mcMMO double drops

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/BlockListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/BlockListener.java
@@ -14,6 +14,7 @@ import com.gmail.nossr50.datatypes.skills.ToolType;
 import com.gmail.nossr50.events.fake.FakeBlockBreakEvent;
 import com.gmail.nossr50.events.fake.FakeBlockDamageEvent;
 import com.gmail.nossr50.events.fake.FakeEvent;
+import com.gmail.nossr50.events.items.McMMOItemSpawnEvent;
 import com.gmail.nossr50.mcMMO;
 import com.gmail.nossr50.skills.alchemy.Alchemy;
 import com.gmail.nossr50.skills.excavation.ExcavationManager;
@@ -111,7 +112,7 @@ public class BlockListener implements Listener {
         //If there are more than one block in the item list we can't really trust it and will back out of rewarding bonus drops
         if (blockCount <= 1) {
             for (Item item : event.getItems()) {
-                ItemStack is = new ItemStack(item.getItemStack());
+                ItemStack is = item.getItemStack();
 
                 if (is.getAmount() <= 0) {
                     continue;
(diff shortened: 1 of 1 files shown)


SrBedrock:
@nossr50 Hi, I’d like to point out an issue related to the way `mcMMO` handles treasure generation using the `BlockDropItemEvent`.  

Currently, the event is registered with **priority HIGHEST**, and the treasure logic depends on the list of block drops being intact. This creates a problem for other plugins that also need to modify or even clear items from that list.  

For example, an auto-pickup plugin uses the same event to check whether the player can collect the drops, and if allowed, removes the items from the drop list. When this happens, mcMMO is no longer able to generate treasures, since it depends on those drops still being present in the list.  

In practice, this means mcMMO is preventing other plugins from safely altering the drop list without breaking treasure generation. It might be better if mcMMO could either:  
- use a lower event priority, or  
- decouple treasure generation from the actual drop list state, so that plugins modifying or removing drops don’t unintentionally block treasure generation.  

This would make mcMMO more compatible with auto-pickup and similar plugins.

src/main/java/com/gmail/nossr50/listeners/BlockListener.java
```diff
@@ -134,17 +135,23 @@ public void onBlockDropItemEvent(BlockDropItemEvent event) {
...
                     final Location centeredLocation = getBlockCenter(event.getBlock());
-                    for (int i = 0; i < bonusCount; i++) {
+                    int newAmount = is.getAmount() * (bonusCount + 1);
 
-                        ItemUtils.spawnItemNaturally(event.getPlayer(),
-                                centeredLocation, is, ItemSpawnReason.BONUS_DROPS);
+                    ItemStack copy = is.clone();
+                    copy.setAmount(newAmount);
+
+                    final McMMOItemSpawnEvent spawnEvent = new McMMOItemSpawnEvent(centeredLocation,
```
nossr50, line 149: I'm not sure it makes sense for this event to fire if all we are doing is modifying the amount of the itemStack in this event. Instead, maybe we need a new event, something like McMMOAddBonusDrops, and it shows the items, the number of items that will be added, the final result, and is cancellable. If it goes uncancelled, or modified, the result affects the outcome of this listener.

nossr50 (changes requested):

nossr50:
> @nossr50 Hi, I’d like to point out an issue related to the way `mcMMO` handles treasure generation using the `BlockDropItemEvent`.
> 
> Currently, the event is registered with **priority HIGHEST**, and the treasure logic depends on the list of block drops being intact. This creates a problem for other plugins that also need to modify or even clear items from that list.
> 
> For example, an auto-pickup plugin uses the same event to check whether the player can collect the drops, and if allowed, removes the items from the drop list. When this happens, mcMMO is no longer able to generate treasures, since it depends on those drops still being present in the list.
> 
> In practice, this means mcMMO is preventing other plugins from safely altering the drop list without breaking treasure generation. It might be better if mcMMO could either:
> 
> * use a lower event priority, or
> * decouple treasure generation from the actual drop list state, so that plugins modifying or removing drops don’t unintentionally block treasure generation.
> 
> This would make mcMMO more compatible with auto-pickup and similar plugins.

I see the problem, we shouuld probably adjust priority to Lowest, change the event to reporting mcMMO adding item quantity, and that would satify the requirements.

nossr50:
feel free to tag me in discord if you want to discuss or ask questions

PR #5217, opened 2025-09-15
Add null-safety and helper methods to AbilityAPI

BitByLogics:
While using the API, I noticed occasional NPEs from the `AbilityAPI`. While I could add null checks myself prior to calling these methods, I felt it was more the API’s responsibility to handle this safely.

# Add null-safety and helper methods to AbilityAPI

This refactor improves the `AbilityAPI` by adding null checks around `UserManager.getPlayer(player)` calls to prevent potential NPEs.

### Changes
- Centralized ability checks into `hasAbilityEnabled(Player, SuperAbilityType)`
- Centralized cooldown setters into `setAbilityCooldown(Player, SuperAbilityType, long)`
- All public methods now safely handle cases where `McMMOPlayer` is not loaded

### Why
Without these checks, calling `AbilityAPI` methods too early (e.g., right after a player joins) can result in NPEs.  
This change makes the API safer to consume for plugins without requiring additional null checks from plugin authors.

### DIFF
--- a/src/main/java/com/gmail/nossr50/api/AbilityAPI.java
+++ b/src/main/java/com/gmail/nossr50/api/AbilityAPI.java
@@ -12,35 +12,39 @@ public final class AbilityAPI {
     }
 
     public static boolean berserkEnabled(Player player) {
-        return UserManager.getPlayer(player).getAbilityMode(SuperAbilityType.BERSERK);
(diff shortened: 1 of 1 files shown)


nossr50:
Thanks, the classes in the API package are some of the oldest and most neglected, I appreciate it.

Issue #5171, opened 2025-04-13
Feature request: Placeholder of the leaderboard.


hudsonandres:
A placeholder for the leaderboard could be added for player at level X for skill Y, this would make it easier to create menus to check an overall ranking, something very detailed for the server members.

Something similar to the image below.
![Image](https://github.com/user-attachments/assets/b2af2bc4-687a-4191-a293-07131beb96ef)

Something similar to the image below.

Additionally, a placeholder can also be added to check if you already have the skill level required to unlock the skill.

Nothing too fancy, something like "%mcmmo_checklevel_<skill>_50%" which returns ✔ or ✘ only.

PR #5229, opened 2025-11-29
Adds leaderboard support to PlaceholderAPI

hudsonandres:
- Adds placeholders to display mcMMO leaderboard information through PlaceholderAPI.

- `%mcmmo_mctop_<skill>:<position>%`
- `%mcmmo_mctop_overall:<position>%`
- `%mcmmo_mctop_name_<skill>:<position>%`
- `%mcmmo_mctop_name_overall:<position>%`
- `%mcmmo_checklevel_<skill>:<level>%`

- Added a new explanatory file `PLACEHOLDERS.md` detailing all the placeholders that can be used.

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/BlockListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/BlockListener.java
@@ -242,11 +242,16 @@ public class BlockListener implements Listener {
             return;
         }
 
-        BlockState blockState = event.getNewState();
+        final BlockState newState = event.getNewState();
+
+        if (!newState.isPlaced()) {
+            // not backed by a real block
+            return;
+        }
 
         if (ExperienceConfig.getInstance().isSnowExploitPrevented() && BlockUtils.shouldBeWatched(
-                blockState)) {
-            Block block = blockState.getBlock();
+                newState)) {
+            final Block block = newState.getBlock();
 
             if (BlockUtils.isWithinWorldBounds(block)) {
                 BlockUtils.setUnnaturalBlock(block);
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -1,3 +1,6 @@
+Version 2.2.050
+    Further improved compatibility with Excellent Enchants
+
 Version 2.2.049
     Combat abilities work with spear in off-hand again (see notes)
     Sweet berry bushes now work with Herbalism (thanks dnocturne)


src/main/java/com/gmail/nossr50/listeners/BlockListener.java
nossr50: - [ ] This seems like it was included in your PR by accident, please remove and if needed create a separate PR for it.

PLACEHOLDERS.md
nossr50: I suppose I'm not against having a readme for placeholders in the repo, but wouldn't updating the wiki make more sense and link to that instead?

nossr50 (changes requested):
Looks like some stuff was included by accident that should be removed.

nossr50:
- [ ] Unit tests would probably be a good idea, you can mock up the responses from the DatabaseManager(s) and just validate the business logic here

src/main/java/com/gmail/nossr50/placeholders/CheckLevelPlaceholder.java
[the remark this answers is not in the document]
hudsonandres: This placeholder would be used to indicate whether you have a high enough level to unlock sub-skills, nothing more than a way to create a custom information menu, like the one in mcMMO GUI. ([https://github.com/mcMMO-Dev/mcMMO/issues/5171](url))

image

src/main/java/com/gmail/nossr50/listeners/BlockListener.java
hudsonandres: Yes, it was for a test I was doing and I included it incorrectly. I've already removed it.

PLACEHOLDERS.md
hudsonandres: Yes, it could be... I had actually created the file to detail the new placeholders, and to make it "more complete" I added the ones that could already be used, but yes, it would be even better to have a link directing straight to the Wiki where it shows all the ones that can be used.

hudsonandres:
> Why is a colon (`:`) used to separate the position instead of an underscore (`_`), which is the commonly adopted standard for placeholders in general?

I only kept the position structure from the original standard. PapiExpansion has a line for that: `int dataPosition = params.indexOf(":");`, however, it can be changed to underscore without any problem; a few adjustments are all that's needed to make it work.

PR #5230, opened 2025-11-30
Update Italian Translation

Tig3r98:
- Added missing translation keys
- File structure now matches locale_en_US


nossr50:
Thank you

Issue #5205, opened 2025-08-10
Add proper support for modded items to mcMMO


nossr50:
Historically, mcMMO has had support for modded items. However, this support for modded items is out of date and has been removed for being non-functional. New code to support modded items should be added.

**Requirements**

- Support and recognize modded items such as custom Swords, Axes, etc
- Support repair/salvage on modded items
- Support being able to add modded items as treasures to Fishing/Excavation/etc

Konpaku-LVX:
@nossr50 i might have found a hint about this? i was playing my arclight server and i noticed that the copper doors for this mod https://modrinth.com/mod/backport-copper-age actually mcMMO is detecting them just fine? when i right click the door it doesnt say "ready your fists", mining the copper ore on 1.21.1 is giving me xp just fine!!!

i think its because its using the minecraft: namespace, and perhaps mcMMO only registers these minecraft: namespaces.

for some reason i feel this could be a huge hint to getting it to work properly!

all it has to do is see the other namespaces somehow? or maybe a compatibility layer between the two

Konpaku-LVX:
actually @nossr50  you might not even need to worry about this in arclights case long term, this might be fixable inside of arclight itself, but i will keep you posted 🫡

Konpaku-LVX:
actually, it seems theres no need to fix anything in arclight. i did an attempt myself, and at least for aether it works. i just defined them manually and made them into an addon plugin. i can release it at some point once i get it polished up some more.

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/4589ce5f-fc5a-43be-a918-299b94591992" />

im not sure how you could make this automatically work with every mod though. it seems like it makes more sense to define them manually since ores and stuff and pickaxe tools etc. might have different durability and XP and stuff

Konpaku-LVX:
once i get it polished up though i could release it and maybe based off that you could find a way to easily let people define their own items and xp and such inside mcMMO.

ecoterrorist:
> once i get it polished up though i could release it and maybe based off that you could find a way to easily let people define their own items and xp and such inside mcMMO.

Just following up with this thread since I also run Arclight servers with mcMMO- is there any publicly available solution for this yet?

Konpaku-LVX:
honestly i havent worked on it much but i can share the files later, need
to sleep now. only set it up for aether but the source code should be
understandable

PR #5239, opened 2025-12-20
feat: add custom block registry for third-party plugin integration

Th0rgal:
## Summary

Hi!

I'm the maintainer of Oraxen and I propose this integration that would make it easier for custom block plugins to work with mcMMO.

This PR adds a new `CustomBlockRegistry` that allows plugins like Oraxen (but also others like ItemsAdder) to register their custom blocks so that players can earn mcMMO skill XP when breaking them.

### What's included:

- **CustomBlockRegistry**: A registry where plugins can register their custom blocks with skill type and XP values
- **McMMOCustomBlockBreakEvent**: An event that fires when a registered custom block is broken, allowing plugins to modify XP dynamically
- **ExperienceAPI additions**: Simple methods for plugin developers to register and unregister their custom blocks

### How it works

Plugin developers can register their custom blocks in their onEnable:

```java
// Register a custom ore to give Mining XP
ExperienceAPI.registerCustomBlock("myplugin", "mythril_ore", "MINING", 150);

// Or using the combined format
ExperienceAPI.registerCustomBlock("myplugin:diamond_ore_variant", "MINING", 200);
```

And clean up in onDisable:

```java
ExperienceAPI.unregisterAllCustomBlocks("myplugin");
```

### Why this matters

Currently, custom block plugins have no clean way to integrate with mcMMO. Players who use both Oraxen and mcMMO often ask why breaking custom ores doesn't give them Mining XP. This PR solves that problem!

### Companion PR

I've also submitted a companion PR to Oraxen that uses these new APIs: https://github.com/oraxen/oraxen/pull/1703

### Backwards compatibility

This is purely additive and doesn't change any existing behavior. All existing blocks and XP handling continues to work exactly as before.

## Test plan

- [ ] Build mcMMO and verify it compiles
- [ ] Test registering a custom block via the API
- [ ] Test that XP is awarded when the custom block is broken
- [ ] Test unregistering blocks
- [ ] Test that existing vanilla block XP still works normally

Thanks for considering this! I'm happy to make any adjustments based on your feedback. Looking forward to better integration between our plugins!

### DIFF
--- a/src/main/java/com/gmail/nossr50/api/ExperienceAPI.java
+++ b/src/main/java/com/gmail/nossr50/api/ExperienceAPI.java
@@ -1320,6 +1320,11 @@ public final class ExperienceAPI {
      * integrate with mcMMO's skill system. Once registered, players will
      * automatically receive XP when breaking the custom block.
      * <p>
+     * <b>Config Integration:</b>
+     * On first registration, an entry is created in {@code custom-blocks.yml} with the
+     * provided default values. Server owners can then customize XP values or disable
+     * specific blocks. Config values always take priority over API-registered values.
+     * <p>
      * Example usage:
      * <pre>
      * // In your plugin's onEnable:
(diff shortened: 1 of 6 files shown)


Th0rgal:
> I'm not opposed to adding an API for an easy "it just works" experience for users of mcMMO and other plugins, however, I think config should be involved in this process.
> 
> If a plugin is registering a custom block, it should make an entry into the relevant config and users should be able to override that entry (for example, setting the XP values to whatever they want).
> 
> So, if a plugin were to register a custom block for a user, it should create relevant experience.yml values if it's the "first contact" of this custom block/etc.
> 
> Perhaps experience related to custom things could simply go in its own config to keep things segregated, it doesn't really matter much, or rather I haven't given it sufficient thought.
> 
> I haven't had time to sit and think about all the things in this PR, this is just my initial thoughts after skimming the files changed.
> 
> I'm reminded of #5205 and #5205, it is likely the work related to those issues would also bring about support for custom blocks/entities/etc.

Thanks for the feedback! I updated the pr to use a config-backed approach. You are right, that should be better.
Changes made:
- Added custom-blocks.yml config file for user customization
- Plugin registrations now create config entries on "first contact" only
- Config values always take priority over API-registered values (users can override XP)
- Added enabled: false option to disable specific blocks
- Custom blocks persist across restarts

The config is now the source of truth but the API stays simple for plugin developers.

Th0rgal:
Got it @nossr50 , thanks for the reply. Let me know what you prefer, I'm open to updating this pr if this helps

PR #5258, opened 2026-02-06
Set max display of spear master to two decimals

Momshroom:
Fix display of spear mastery values.  Floating point issues were causing it to show with 16 decimals at some levels.


nossr50:
Thank you

PR #5265, opened 2026-02-28
Fix stacked Chimaera Wing item loss

UDPSendToFailed:
Before: Having more than one Chimaera Wing in the main hand and right clicking to use them deletes the entire stack.

After: It removes one Chimaera Wing from the stack as it's supposed to.


nossr50:
Thank you!

PR #5272, opened 2026-04-17
Make crossbow arrow with potion type a tipped arrow

### LINKED ISSUE #5243: Trickshot Ability Broken On Tipped Arrows
Server Version: Paper 1.21.8
McMMO Version: mcMMO-2.2.048

When the arrow bounces from the crossbow ability, if the arrow is tipped (i.e. with speed), the tipped arrow:
- Loses its texture and will look like a normal arrow
- The duration of the effect is 8x longer

Example:
I shoot a crossbow with a tipped arrow that has speed (00:22) duration. The arrow bounces. I go and pick up the arrow. Instead of the arrow looking blue (for speed), it looks like a normal arrow. Along with that, the arrow now has a (03:00) duration for speed.

Here's the link to a video: [https://streamable.com/x2q68v](https://streamable.com/x2q68v)

In an actual pvp scenario, you can get 8x longer duration poison arrows and use to to get a significant advantage.

**Note**
When testing, make sure you're in **survival mode**. I was trying to record the video for the demo and it won't work if you're in creative.

### LINKED ISSUE #5246: Trickshot makes tipped arrows OP
Tipped arrows shot from a crossbow that ricochet off of an object from trickshot results in arrows that are still tipped, but the time of the potion affects aren't correct, it goes to the splash potion's duration.

For example, a ricocheted Poison II (2 seconds) tipped arrow results in a poison II arrow with a duration of 21 seconds. Slowness goes from 30 seconds to 4 minutes. 

mcMMO version 2.2.045
Paper version 1.21.10-129

flyncodes:
Fix #5243 and fix #5246

I have tested this and it fixed the issue for me.

### DIFF
--- a/src/main/java/com/gmail/nossr50/skills/crossbows/CrossbowsManager.java
+++ b/src/main/java/com/gmail/nossr50/skills/crossbows/CrossbowsManager.java
@@ -79,7 +79,7 @@ public class CrossbowsManager extends SkillManager {
             spawnedArrow.setItem(new ItemStack(Material.TIPPED_ARROW));
             spawnedArrow.setBasePotionType(originalArrow.getBasePotionType());
         }
-        // copy some properties from the old arrow-
+        // copy some properties from the old arrow
         spawnedArrow.setShooter(originalArrowShooter);
         spawnedArrow.setCritical(originalArrow.isCritical());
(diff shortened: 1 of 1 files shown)


nossr50:
Thank you!

PR #5277, opened 2026-05-01
Fix potion lengths being shorter/longer for newer potion types than vanilla

Warriorrrr:
Corrects the default potion durations for splash/lingering oozing, weaving, infested & wind charged potions. The splash duration for these is the same length as the normal duration, and the lingering duration is only 45 seconds.


nossr50:
Thank  you!

PR #5282, opened 2026-05-08
Add option to toggle receiving conflicting enchants from magic hunter

Warriorrrr:
This PR adds a toggle for whether magic hunter can give you items with conflicting enchantments, such as multiple types of protection on a single armor piece.

Depending on who you ask this is either a bug or a feature, so the safest bet is to have a config option for it. The proposed default here is to have it off (leaning towards this being a bug rather than a feature), but if that isn't the case then I'll happily change that default here.


nossr50:
Ah thanks for this @Warriorrrr , will review soon

nossr50:
Thank you!

PR #5283, opened 2026-05-09
Fix excavation treasures not dropping

Warriorrrr:
Fixes a bug from yesterday's commit https://github.com/mcMMO-Dev/mcMMO/commit/ce021dbe129c912820f28649d6c4050ed80abe13 that caused excavation treasures to not work anymore. The problem was that in the block drop item event the block is already air, so affectedByGigaDrillBreaker would always be false and the treasure drops would also be empty if that were to pass. The solution to this is to use the block state provided by the event, which does reflect the block that used to be there.


Warriorrrr:
Solved by https://github.com/mcMMO-Dev/mcMMO/commit/47cebf2cc2830eac69af3e4e5cd0d20cf9d81c4e

nossr50:
Sorry I didn't even see your PR @Warriorrrr , I found this in my own testing and ended up resolving it already 🫡

Warriorrrr:
All good, you also managed to catch the issue with metadata which I didn't. Thank you for your work!

PR #5294, opened 2026-05-29
Fix traveling block metadata leaking on folia

Warriorrrr:
On folia, entity tasks are not executed when an entity is removed, which caused this metadata to stick around forever. This fixes that by adding a callback to the task that runs when the entity is removed to also clean up the metadata.


nossr50:
Thank you!

PR #5295, opened 2026-05-29
Fix entities keeping their health bars after unload

Warriorrrr:
Fixes an issue where you could hit an entity to trigger the health bar, unload the chunk (e.g. by disconnecting), and the entity would retain the health bar indefinitely.


nossr50:
Thank you!

PR #5298, opened 2026-06-03
Refactor scoreboards to packet-based impl for Folia/Canvas compatibility

albert4719:
Replace Bukkit Scoreboard/Objective API with scoreboard-library (packet-based)
  to support Folia and Canvas. Key changes:

  - ScoreboardWrapper: sidebar now uses packet-based Sidebar; remove/revert board is sidebar.removePlayer() instead of player.setScoreboard()
  - ScoreboardManager: load ScoreboardLibrary in init() with NoopScoreboardLibrary fallback; randomize POWER_OBJECTIVE name per startup to avoid colliding with leftover objectives from the old Bukkit impl on client reconnect
  - ScoreboardManager: fix getWrapper() leaking Sidebar resources when wrapper was not stored in map; fix double task cancel in teardownPlayer/cleanup
  - ScoreboardManager: call removeScore() on player quit so offline players do not accumulate in the scores map and flood new players with a large initial packet burst
  - ScoreboardManager: guard removePlayer() with isOnline() to avoid NPE in scoreboard-library's async tick (weak-key playerMap GC race)
  - McMMOScoreboardMakeboardEvent and McMMOScoreboardRevertEvent are still fired for API compatibility; McMMOScoreboardObjectiveEvent is deprecated (no Bukkit Objective exists in the packet impl)
  - pom.xml: fix shade includes for scoreboard-library (scoreboard-library-api + scoreboard-library-implementation); add ViaVersion to softdepend

### DIFF
--- a/src/main/java/com/gmail/nossr50/util/scoreboards/ScoreboardManager.java
+++ b/src/main/java/com/gmail/nossr50/util/scoreboards/ScoreboardManager.java
@@ -11,33 +11,34 @@ import com.gmail.nossr50.locale.LocaleLoader;
 import com.gmail.nossr50.mcMMO;
 import com.gmail.nossr50.util.LogUtils;
 import com.gmail.nossr50.util.Misc;
+import com.gmail.nossr50.util.PaperUtil;
 import com.gmail.nossr50.util.player.UserManager;
+import com.gmail.nossr50.util.scoreboards.backend.BukkitScoreboardBackend;
+import com.gmail.nossr50.util.scoreboards.backend.NoopScoreboardBackend;
+import com.gmail.nossr50.util.scoreboards.backend.PlayerBoard;
+import com.gmail.nossr50.util.scoreboards.backend.PacketScoreboardBackend;
+import com.gmail.nossr50.util.scoreboards.backend.ScoreboardBackend;
+import com.gmail.nossr50.util.scoreboards.backend.ScoreboardBackendSelector;
+import com.gmail.nossr50.util.scoreboards.backend.ScoreboardBackendType;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import net.megavex.scoreboardlibrary.api.objective.ScoreboardObjective;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.entity.Player;
-import org.bukkit.scoreboard.DisplaySlot;
-import org.bukkit.scoreboard.Objective;
 import org.bukkit.scoreboard.Scoreboard;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Manages the Scoreboards used to display a variety of mcMMO related information to the player
  */
 public class ScoreboardManager {
-    static final Map<String, ScoreboardWrapper> PLAYER_SCOREBOARDS = new HashMap<>();
-
-    // do not localize; these are internal identifiers
-    static final String SIDEBAR_OBJECTIVE = "mcmmo_sidebar";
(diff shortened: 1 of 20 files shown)


nossr50:
@albert4719 take a look at the recent changes, I modified it so that it should only deploy this new API on Folia-based servers.

albert4719:
> @albert4719 take a look at the recent changes, I modified it so that it should only deploy this new API on Folia-based servers.

LGTM. BTW, packet backend shows up fine on Paper too, not just Folia.

The only difference is the Bukkit path still fires `McMMOScoreboardObjectiveEvent` and the packet path skips it.

Works as expected on my end.

src/main/java/com/gmail/nossr50/util/scoreboards/backend/BukkitScoreboardBackend.java
```diff
@@ -0,0 +1,140 @@
...
+import org.bukkit.entity.Player;
+import org.bukkit.scoreboard.DisplaySlot;
+import org.bukkit.scoreboard.Objective;
+import org.bukkit.scoreboard.Scoreboard;
+import org.bukkit.scoreboard.ScoreboardManager;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class BukkitScoreboardBackend implements ScoreboardBackend {
+    private static final String POWER_OBJECTIVE = "mcmmo_" + java.util.UUID.randomUUID()
```
albert4719: One issue here: `POWER_OBJECTIVE` uses a random UUID-based name. That makes sense for the packet impl — regenerated each startup to avoid clashing with existing objectives. But the Bukkit path shouldn't reuse this;

albert4719: The original Bukkit value was `mcmmo_pwrlvl` — best to change it back for compatibility, since this objective is persisted to the world's scoreboard data and survives restarts.

albert4719: Oh I just noticed there's an unregister() here already, my bad. 
                                                                                                                                                                                                                                                                                                                                                                                                           But I'd still go with the fixed `mcmmo_pwrlvl` instead of a random name. Servers coming from the old mcMMO already have mcmmo_pwrlvl saved in their scoreboard.dat, so reusing it keeps things tidy instead of leaving the old one orphaned. And the unregister only helps on a clean shutdown — if the server crashes it won't run, so a random name leaves an orphan behind, and every restart adds a new one. A fixed name just reuses the same objective so that never happens

src/main/java/com/gmail/nossr50/util/scoreboards/ScoreboardManager.java
```diff
@@ -11,33 +11,43 @@
...
 
     // do not localize; these are internal identifiers
-    static final String SIDEBAR_OBJECTIVE = "mcmmo_sidebar";
-    static final String POWER_OBJECTIVE = "mcmmo_pwrlvl";
+    public static final String SIDEBAR_OBJECTIVE = "mcmmo_sidebar";
+    // Randomized per server start so it can never collide with a leftover objective of the same
+    // name on a client - e.g. the old Bukkit implementation's 'mcmmo_pwrlvl' lingering after an
+    // in-place upgrade, or another plugin's objective. Minecraft caps objective names at 16 chars,
+    // so this stays well under: "mcmmo_" (6) + 8 hex = 14.
+    public static final String POWER_OBJECTIVE = "mcmmo_" + java.util.UUID.randomUUID()
```
albert4719: These three constants (POWER_OBJECTIVE, SIDEBAR_OBJECTIVE, DISPLAY_NAME) seem unused now that each backend defines its own objective names — looks safe to remove

nossr50:
@albert4719 run into any issues with this or has it been smooth sailing?

albert4719:
@nossr50 Just submitted a couple review comments — I'd written them a
while back but forgot to actually submit the review, so they were
stuck in pending on my side. Sorry for the delay.

nossr50:
@albert4719 try this version out

albert4719:
@nossr50 Tested on both Paper and Canvas — scoreboard renders correctly and 
everything's working smoothly.  LGTM 👍

nossr50:
Thanks for this @albert4719

PR #5306, opened 2026-07-04
gitleaks hits old Jenkins marker in history (optional purge script)

smmariquit:
`master` is clean. Metrics are long gone.

gitleaks on `--all` still flags a hex string from 2013-era metrics code. It is not a bStats API key. It was a hardcoded Jenkins build fingerprint (`officialKey` compared to `.jenkins`) so old graphs could report official CI builds.

No rotation needed. Low priority.

This PR adds a short note + `scripts/purge-metrics-official-key-history.sh` if you ever want to scrub it from history for scan hygiene. I cannot force-push your remote.

`.md` / `.sh` are force-added because your `.gitignore` ignores them.

### DIFF
--- a/docs/security/metrics-official-key-git-history.md
+++ b/docs/security/metrics-official-key-git-history.md
@@ -1,0 +1,20 @@
+# Old Jenkins build marker in git history
+
+`master` is fine. Metrics code is gone.
+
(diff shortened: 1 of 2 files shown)


smmariquit:
Closing this. On reflection the history scrub isn’t worth the noise for a dead 2013 Jenkins marker. master is already fine. Sorry for the drive-by.

PR #5310, opened 2026-07-06
Move fish exploit checks to the lowest priority

FireML:
Moves fishing exploit checks to the lowest event priority to allow third party plugins to properly check for exploits.

It currently acts like so: The EvenMoreFish plugin checks for exploits, sets a custom item, sends a chat message, and then mcMMO performs its checks and removes the item.


nossr50:
A few things need to change before this can be merged:

**1. The handler isn't actually at LOWEST priority.** A bare `@EventHandler` defaults to NORMAL (https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/EventHandler.html). It needs `@EventHandler(priority = EventPriority.LOWEST)`. Worth noting that even at LOWEST this only guarantees running before higher-priority handlers; if another plugin also listens at LOWEST, ordering between the two is registration order (https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/EventPriority.html). LOWEST is still the earliest slot available.

**2. Removing the early return means exploit-flagged catches still pay mcMMO rewards.** That `return` was the only thing gating rewards; `processFishing` doesn't check `isExploitingFishing()` itself: https://github.com/mcMMO-Dev/mcMMO/blob/3a711577530dc539787bfaf68770717b515018f1/src/main/java/com/gmail/nossr50/skills/fishing/FishingManager.java#L404-L486 With this change, a flagged catch loses the vanilla item and XP but still gets full Fishing XP, a Treasure Hunter roll, and an ExtraFish spawn from the later handler. Add a guard in the CAUGHT_FISH case before `processFishing`, e.g. `if (ExperienceConfig.getInstance().isFishingExploitingPrevented() && fishingManager.isExploitingFishing()) return;`. Handlers for one event run in priority order on the same thread, so the LOWEST handler's `processExploiting` call has always completed before this guard reads the result.

**3. The exploit check now runs for more than actual fish catches.** In the old code this check only ran inside `case CAUGHT_FISH:` of the switch in the monitor handler. The new handler instead filters with `event.getCaught() instanceof Item`, which is not equivalent: PlayerFishEvent also fires with state CAUGHT_ENTITY when a player reels in an entity with the rod, and if that entity is a dropped item floating in the water, `getCaught()` returns an Item there too (https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/PlayerFishEvent.State.html). So as written, reeling in dropped items counts toward the overfish limit, and once the player is flagged the handler deletes the hooked item, which could be another player's dropped loot. Add `if (event.getState() != PlayerFishEvent.State.CAUGHT_FISH) return;` at the top of the new handler to keep the check scoped the way it was.

**4. Use `ListenerGuards.resolveEligiblePlayer` instead of the hand-rolled guard block.** Master centralized the world blacklist / WorldGuard / profile checks; see the other fishing handlers: https://github.com/mcMMO-Dev/mcMMO/blob/3a711577530dc539787bfaf68770717b515018f1/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java#L424-L431.

nossr50 (changes requested):
see prior comment

FireML:
I made lots of basic mistakes here, this is why we don't code while tired. I'll try to resolve the issues later today 🙂

FireML:
All issues should be resolved :slightly_smiling_face:

nossr50:
@FireML looks like your diff includes some stuff by mistake, mind cleaning it up and then I'll review?

FireML:
i can't get the hang of this so i'll open a new PR...

PR #5313, opened 2026-07-09
Move fish exploit checks to the lowest priority

FireML:
See #5310

### DIFF
--- a/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
+++ b/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java
@@ -361,6 +361,9 @@ public class PlayerListener implements Listener {
                     .getFishingExploitingOptionMoveRange()));
             event.setExpToDrop(0);
             caughtItem.remove();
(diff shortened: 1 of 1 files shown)


nossr50:
We want to keep the current deterrent behavior (item removed, no XP) as is. Two things before this is mergeable:

**The spam fishing check has the same problem and didn't move.** The `isFishingTooOften()` branch still zeroes XP and removes the item at HIGHEST: https://github.com/mcMMO-Dev/mcMMO/blob/980b6cbe9953b2c47da4df976a817d0ff8ac21dd/src/main/java/com/gmail/nossr50/listeners/PlayerListener.java#L430-L442 so the scenario from #5310 still happens through that path. Both denials should live in the new LOWEST handler.

**Gate the new handler on the exploit fix config first.** The denial itself is already inside the `isFishingExploitingPrevented()` check, but the handler resolves the mcMMO player and runs permission checks before ever consulting it. Servers that disabled `ExploitFix.Fishing` in experience.yml shouldn't pay for any of that on every catch, so bail on the config check at the top of the handler.

PR #5314, opened 2026-07-09
Fix NPE if debug command is used before profile load

Warriorrrr:
A small thing that's caused exceptions a couple of times for me due to a player using a mod to run this command right after joining.


nossr50:
@Warriorrrr thank you, appreciate all your contributions 🫡

PR #5332, opened 2026-08-13
refactor: replace Rupture's Bukkit metadata with an owned registry

wling-art:
Rupture's failsafe cancel forgot to clear its entity metadata, so the McMMOPlayer chain behind it never got GC'd — 21% of heap in prod.

65742abe33d582c03485b93add6fcd2d

Moved Rupture to its own registry, RuptureTask tracks and untracks itself. deleted RuptureTaskMeta and the metadata key too, nothing needs them now.


src/main/java/com/gmail/nossr50/mcMMO.java
```diff
@@ -555,6 +555,11 @@ public void onDisable() {
             papiExpansion = null;
         }
         getFoliaLib().getScheduler().cancelAllTasks(); // This removes our tasks
+        for (org.bukkit.World world : getServer().getWorlds()) {
+            for (org.bukkit.entity.Entity entity : world.getEntities()) {
+                entity.removeMetadata(MetadataConstants.METADATA_KEY_RUPTURE, this);
+            }
+        }
```
Warriorrrr: This doesn't seem necessary, if this PR already fixes the leak then cleaning up here isn't needed, and cleaning up metadata during shutdown also isn't needed since it lives in memory only, the server will close shortly after this anyways

Warriorrrr (commented):
I can still see a possibility of this leak happening on folia, when the entity is removed then the task will stop running immediately and won't trigger any of the cleanup logic in RuptureTask. If you could add a retired callback that cleans up the metadata to the place in SwordsManager where the rupture task is started then that would take care of that as well.

src/main/java/com/gmail/nossr50/runnables/skills/RuptureTask.java
```diff
@@ -51,6 +59,11 @@ public RuptureTask(@NotNull McMMOPlayer ruptureSource, @NotNull LivingEntity tar
         this.damageTickTracker = 0;
         this.animationTick = ANIMATION_TICK_INTERVAL; //Play an animation right away
         this.pureTickDamage = pureTickDamage;
+        ACTIVE_RUPTURES.put(targetEntity.getUniqueId(), this);
```
nossr50: A leak is technically possible because you are storing `RuptureTask` in the new concurrent map before anything is registered to remove it.

Example:

1. Another plugin creates a RuptureTask using mcMMO's public constructor.
2. Just creating it adds the victim to mcMMO's "currently bleeding" map.
3. The plugin never starts the task (either because of its own bad code, or hit an error).
4. Every cleanup path lives inside the running task, and this task never runs, so the map entry is never removed.
5. Result: that victim can never be ruptured again until restart, and the map entry keeps the attacker's `McMMOPlayer` (and everything it holds) in memory forever.

I suggest only registering a `RuptureTask` at the same time you register it.

So my suggestion is

1. Delete the put from the constructor, so creating a `RuptureTask` does nothing but create an object.
2. Give `RuptureTask` one new public method, `schedule()`, that adds the task to the bleeding map and then starts it ticking via runAtEntityTimer(targetEntity, this, this::cancel, 1, 1). The this::cancel is for Folia: when the entity dies or unloads, Folia stops its tasks without running them again, and instead calls this callback, so the map entry still gets removed. On Paper/Spigot the callback never fires and the task cleans up itself.
3. In `SwordsManager`, replace the raw scheduler call with ruptureTask.schedule().
4. Leave cancel() exactly as you have it in the PR now


Example of what `schedule()` could be:

```java
/**
 * Registers this rupture as the target's active bleed and starts it ticking.
 * The task removes itself from the registry when it is cancelled; on Folia the
 * retired callback covers entities that are removed before the next tick.
 */
public void schedule() {
    ACTIVE_RUPTURES.put(targetEntity.getUniqueId(), this);
    mcMMO.p.getFoliaLib().getScheduler().runAtEntityTimer(targetEntity, this, this::cancel, 1, 1);
}
```

src/main/java/com/gmail/nossr50/datatypes/meta/RuptureTaskMeta.java
nossr50, line 1: Rather than deleting `METADATA_KEY_RUPTURE` and `RuptureTaskMeta`, keep them and deprecate them so plugins compiling against us don't break.

1. Restore `METADATA_KEY_RUPTURE` in `MetadataConstants` like this:

```java
/**
 * No longer used. mcMMO stopped applying this metadata to bleeding entities in 2.3.001,
 * so {@code hasMetadata} checks against this key always return false. Use
 * {@link com.gmail.nossr50.api.AbilityAPI#isBleeding(LivingEntity)} instead.
 *
 * @deprecated mcMMO no longer applies this metadata; see {@link AbilityAPI#isBleeding}.
 */
@Deprecated(forRemoval = true, since = "2.3.001")
public static final @NotNull String METADATA_KEY_RUPTURE = "mcmmo_rupture";
```

2. Restore `RuptureTaskMeta` with the same treatment:

```java
/**
 * No longer used. mcMMO stopped attaching rupture state to entities as Bukkit metadata
 * in 2.3.001; active ruptures are tracked internally instead. Use
 * {@link com.gmail.nossr50.api.AbilityAPI#isBleeding(LivingEntity)} to check for an
 * active bleed, or {@link RuptureTask#getActive(Entity)} to retrieve the task.
 *
 * @deprecated mcMMO no longer creates or reads this metadata.
 */
@Deprecated(forRemoval = true, since = "2.3.001")
public class RuptureTaskMeta extends FixedMetadataValue {
```

3. Nothing else changes: mcMMO's own code shouldn't reference either one anymore, and taking the key out of `MOB_METADATA_KEYS` is fine as-is since the metadata is never set.

This keeps addons compiling with a warning that points them at the replacement instead of a hard error. Plugins reading the raw `"mcmmo_rupture"` string will still silently see no bleeds. Actual removal of both members can happen in a later release like the rest of the deprecated API.

nossr50 (changes requested):
A few changes and I'd be happy to merge.

nossr50 (approved):

PR #5334, opened 2026-08-16
Blast Mining drops behave like mining with the held pickaxe

RasmusKD:
Debris drops its block item at a flat 10% rate, ignoring the held tool. Plain TNT has dropped 100% since 1.14, so using the skill destroys loot.

Now every block drops what mining it with the held pickaxe would give, at the explosion's own yield. Silk and fortune apply, tool-less loot without a pickaxe, and the tntExplosionDropDecay gamerule keeps working.

The loot tables decide everything, so the manual lists are gone: spawners and budding amethyst still drop nothing, infested gives its regular variant only with silk (reverts the 2.2.032 exclusion, which patched the old fabricating debris path), and BLAST_MINING_BLACKLIST could never match ore drops anyway.

Silk blasts also remove their blocks from the explosion and break them in the handler: the explosion pops block XP tool-lessly, and silk plus that XP was a free loop on re-placed ores. Placed blocks drop but earn no XP or bonuses. TNT stays with the explosion so chains prime.

Ore bonuses and the yield ceiling are untouched. Tested live: silk gives blocks and zero XP orbs, spawners and budding stay empty, infested gives stone, chained TNT primes without refunds.

### DIFF
--- a/src/main/java/com/gmail/nossr50/skills/mining/MiningManager.java
+++ b/src/main/java/com/gmail/nossr50/skills/mining/MiningManager.java
@@ -30,8 +30,6 @@ import com.gmail.nossr50.util.skills.SkillUtils;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import java.util.Locale;
-import java.util.Set;
 import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockState;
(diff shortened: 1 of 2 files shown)


RasmusKD:
Some history for context: debris dropping at nearly full rate was mcMMO's own behavior until mid-2024. Before 2debcbef the debris chance was 'yield - debris reduction', which respected the explosion yield and worked out to roughly 90-100% depending on rank. 2debcbef replaced that with a flat 50% roll, and 25c89c5b (2.2.032) lowered it to 10% as part of the balance pass. Reading the 2.2.032 notes, that pass was about bonus drops and the yield ceiling being too good, which this PR doesn't touch - base debris at 10% mostly just deletes cobble a plain TNT blast would have dropped.

RasmusKD:
Went full vanilla parity here: the lists were patching the old debris path, which fabricated block items without consulting loot tables. Now loot decides everything: spawners and budding drop nothing, infested gives its regular variant only with silk. That reverts the 2.2.032 exclusion, easy to re-add if you want it. BLAST_MINING_BLACKLIST also went, ore drops could never contain its materials.

nossr50:
I'll have to think about this one, the main thing is that people find the skill sort of OP, so I have to be careful when making changes that will adjust its effectiveness.

PR #5335, opened 2026-08-16
Catalysis updates the real brewing stand instead of a snapshot

RasmusKD:
AlchemyBrewTask keeps the BlockState it was created with and calls setters on it every tick, but snapshot setters never reach the world. The visible effects:

- with Catalysis on vanilla ingredients, the client's brew arrow follows vanilla's own timer, so the potion finishes while the arrow is still mid-bar
- custom-ingredient brews show no progress at all, the potion just pops after x seconds with no feedback
- the initial fuel decrement for custom ingredients is lost
- the brewing-time reset in cancelBrew() does nothing

Writes now go through a fresh state fetched and applied within the same tick, so the arrow tracks the accelerated brew and fuel/cancel actually happen. One method added, three call sites changed.

### DIFF
--- a/src/main/java/com/gmail/nossr50/runnables/skills/AlchemyBrewTask.java
+++ b/src/main/java/com/gmail/nossr50/runnables/skills/AlchemyBrewTask.java
@@ -143,7 +143,7 @@ public class AlchemyBrewTask extends CancellableRunnable {
     private void initializeBrewing() {
         if (firstRun) {
             firstRun = false;
-            ((BrewingStand) brewingStand).setFuelLevel(fuel);
+            applyToLiveStand(stand -> stand.setFuelLevel(fuel));
         }
     }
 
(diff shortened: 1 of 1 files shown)


nossr50 (changes requested):
The fuel write in `AlchemyBrewTask.initializeBrewing()` causes a bug now that it actually works. It puts back fuel that vanilla already spent:

1. When the ingredient goes in, the `AlchemyBrewTask` constructor reads the fuel level from its snapshot. At that moment the brew hasn't started, so it reads the full amount.
2. A tick later, vanilla starts the brew in `BrewingStandBlockEntity.serverTick` and spends one fuel.
3. On the task's first run, `initializeBrewing()` writes the old number back. The spent fuel comes back.

I tested this on a Paper 26.2 server with some debug logging around the write. The stand had 20 fuel, I brewed nether wart, and the log showed the stand at 19 right before the write and 20 right after. Same thing on every nether wart brew I did, and after several full brews the stand still had all 20 fuel. So with this PR, brewing vanilla ingredients on an owned stand costs no fuel.

This might fix it: delete the fuel write, plus the fuel capture and the `== -1` check in the constructor that go with it. Vanilla already handles fuel for its own ingredients. The `== -1` check doesn't do anything anymore anyway, an idle stand reports 0 these days, not -1. Custom ingredients not using fuel is a different, older problem. Open to other ideas too.

RasmusKD:
Good catch. The write was dead while it targeted the snapshot, so the stale read never mattered until this PR made it live. Removed the fuel write, the capture and the -1 check as suggested.

nossr50:
I'm still considering some stuff regarding this change, but it will probably come in soon.
Thanks for the PR.

nossr50:
Sorry don't mean to leave you hanging on this, overall it's a good change, I've just had a very busy month and this month is looking equally chaotic. I just had some perf concerns with this approach that I wanted to address before any merge (or rather, confirm to be a non-issue).

PR #5343, opened 2026-09-16
Fix FlatFile database truncation on read failure

cyrne1-7208:
## Summary
Fixes FlatFile database (`mcmmo.users`) truncation and data loss when an `IOException` occurs while reading the database file.

## Problem
In `FlatFileDatabaseManager`, several operations read existing records into an in-memory buffer before writing. Previously, if an `IOException` occurred during the read phase, the methods caught the exception, logged it, and continued straight into their write phases:
1. `newUser()` caught `IOException` during read, logged it, and entered its `FileWriter` block, overwriting the entire database with only the single new player.
2. `purgePowerlessUsers()` caught `IOException` during read and proceeded to call `writeStringToFileSafely()`, overwriting the database with an incomplete buffer.
3. `rewriteUsersFile()` caught `IOException` during read and called `writeStringToFileSafely()`, causing data loss for all callers (`purgeOldUsers()`, `removeUser()`, and `saveUserUUIDs()`).
4. `saveUserUUID()` caught `Exception` during read and proceeded to call `writeStringToFileSafely()`.
5. `checkFileHealthAndStructure()` caught `IOException` during validation, but if any flags had been collected prior to the error, proceeded to overwrite the database file.

## Solution
Guards each operation so that write execution is never reached if reading fails:
- **`newUser()`**: Returns an unloaded `PlayerProfile` immediately upon `IOException`, skipping the `FileWriter` block entirely so existing database content is preserved.
- **`purgePowerlessUsers()`**: Returns `0` immediately upon `IOException`, skipping `writeStringToFileSafely()`.
- **`rewriteUsersFile()`**: Returns `false` upon `IOException` without calling `writeStringToFileSafely()`. Callers (`purgeOldUsers()`, `removeUser()`, `saveUserUUIDs()`) check the return value and abort early on read failure (e.g. `removeUser()` skips `Misc.profileCleanup()` if read failed).
- **`saveUserUUID()`**: Returns `false` immediately upon read exception, skipping `writeStringToFileSafely()`.
- **`checkFileHealthAndStructure()`**: Returns `null` immediately upon validation `IOException`, skipping the update block.

All write mechanisms, exception types, logging formats, and normal-operation semantics remain completely unaltered.

## Testing
- Added unit tests in `FlatFileDatabaseManagerTest` that inject mid-read I/O errors and verify that the database file remains 100% byte-for-byte identical across all operations.
- Ran all 45 tests in `FlatFileDatabaseManagerTest` (all passed).
- Ran full project test suite (`mvn test`, 1,789 tests passed).


cyrne1-7208:
Sorry, I found an issue with this PR and am working on a fix. Please bear with me for a little while.

cyrne1-7208:
> Sorry, I found an issue with this PR and am working on a fix. Please bear with me for a little while.

The issue has been fixed.

PR #5344, opened 2026-09-16
Avoid in-place Vector mutation in Crossbow ricochet calculations

cyrne1-7208:
## Summary

In `CrossbowsManager.handleRicochet()`, Bukkit `Vector` methods (`subtract`, `multiply`) mutate vectors in place. Previously, the reflection formula mutated `arrowInBlockVector` and `normal`, relying on mutation side effects for the subsequent angle check.

This change keeps the calculation side-effect free and adds a regression test ensuring vector arguments are not mutated.

## Changes

- Clone vectors before mutating operations (`subtract`, `multiply`) so `originalArrow.getVelocity()` and the supplied `normal` remain unmodified.
- Simplify the glancing shot angle check to `reflectedDirection.angle(normal) < Math.PI / 4`, removing `inverseNormal` and matching the existing comment.
- Add `CrossbowsManagerTest#handleRicochetShouldNotMutateSuppliedVectors` to guarantee input vectors are not mutated.

## Testing

- `mvn test -Dtest=CrossbowsManagerTest` (14/14 passed).
- Full suite `mvn test` (1,790/1,790 passed).


cyrne1-7208:
This might be a relatively minor cleanup, so please feel free to close this PR if you feel it isn't needed. Thanks!

